PR #5, opened 2017-05-19
Add link to license in manifest

pmlandwehr:
Hey-lo,

I'm [building a version](https://github.com/conda-forge/staged-recipes/pull/2997) of `plotnine` using [`conda`](http://conda.pydata.org/) for [conda-forge](http://conda-forge.github.io/). When possible, we try to include a link to the license file in the `meta.yaml` specification for the build; doing so requires the license be indexed in an explicit [`MANIFEST.in`](https://docs.python.org/2/distutils/sourcedist.html#manifest-related-options) file so that it gets included in the source distribution.

Currently I'm using a local copy of the license, but this pull should add a copy of the license to the next source build.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include versioneer.py
 include plotnine/_version.py
+include LICENSE


has2k1:
@pmlandwehr, thanks for catching that and thank you for the recipe.

Also, two questions
1. Would be be of any benefit if I cut a new release that includes the licence?
2. What is the common practice for package maintainers to make sure that the recipes are up-to-date.

pmlandwehr:
1. You don't need to cut a new version yet, though thanks for the offer. We
can use the local copy that I made. If the license gets rolled into the
next version, that'll be great.

2. Recipes are updated on an ad-hoc basis as maintainers get to them. We
use a few tools to help maintainers keep track of what's changed, and
end-users submit issues and pull requests along for new versions. If you're
interested, I can add you as a maintainer to the plotnine and mizani
recipes.

has2k1:
Yes you can add me as a maintainer, and point me to the tools, maybe I can incorporate it into the release process.

pmlandwehr:
Ok, I've submitted pulls to add you as a a maintainer to https://github.com/conda-forge/mizani-feedstock and https://github.com/conda-forge/plotnine-feedstock. I'll merge the pulls once the CIs have finished - this is unnecessary, but I'm superstitious.  conda-forge's scripts should eventually move you onto the teams.

The key tools I'd point you to are [`feedstockrot`](https://github.com/axiom-data-science/feedstockrot), which checks if feedstocks are out of date relative to pypi, and the [`tick-my-feedstocks.py`](https://github.com/conda-forge/conda-forge.github.io/blob/master/scripts/tick_my_feedstocks.py) script, which checks if feedstocks are out of date and -if so- attempts to update them with the latest content.

*That said*, the structure of the recipe's `meta.yaml` is pretty straightforward - [this](https://github.com/conda-forge/plotnine-feedstock/blob/master/recipe/meta.yaml) is the one for `plotnine`. If you're updating the module's requirements, you'll need to patch the `run:` block. Otherwise, when you update the version all you should need to do is reset the `build`  variable to 0, update the `version` variable to the latest version, and update the `hash` variable to the new sha256 checksum. `tick_my_feedstocks.py` should be able to handle that.

PR #30, opened 2017-07-06
Fix coord_cartesian y-axis scaling

kdauria:
This corrects a bug where y-axis scaling was calculated from the `xlim` argument.


has2k1:
Thanks for the fix.

PR #33, opened 2017-07-08
Fix using classes in stat and geom creation.

ChickenProp:
Speifically, things like `stat_identity(geom=geom_point)` and
`geom_point(position=position_jitter)` now work.
`geom_point(stat=stat_identity)` already worked (624a5cda).

### DIFF
--- a/plotnine/tests/test_geom.py
+++ b/plotnine/tests/test_geom.py
@@ -4,6 +4,7 @@ from __future__ import (absolute_import, division, print_function,
 import pandas as pd
 import pytest
 
+import plotnine as p9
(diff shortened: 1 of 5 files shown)


plotnine/tests/test_geom.py
```diff
@@ -4,6 +4,7 @@
 import pandas as pd
 import pytest
 
+import plotnine as p9
```
has2k1: Can you explicitly import the objects being used in the tests.

ChickenProp:
Sure thing, I've fixed those issues.

has2k1:
Thanks.

PR #35, opened 2017-07-11
Document that labels can take a callable

lorin:
Add some docs that describe that you can pass a callable to the labels parameter.


lorin:
I forgot about the "breaks" argument, I'll add that as well...

has2k1:
Great. Thanks.

PR #51, opened 2017-08-27
Fix dpi option in ggplot.save method

yejianye:
To reproduce the bug, set any dpi when invoking ggplot.save method. It will have no effect.

This is because when dpi is set, it doesn't update self.theme nor adding dpi into `fig_kwargs`. This pull request fixes this bug.

### DIFF
--- a/plotnine/ggplot.py
+++ b/plotnine/ggplot.py
@@ -541,17 +541,13 @@ class ggplot(object):
                 "not pixels). If you are sure you want these "
                 "dimensions, use 'limitsize=False'.".format(width, height))
 
-        # dpi
         if dpi is None:
             try:
-                dpi = self.theme.themeables.property('dpi')
+                self.theme.themeables.property('dpi')
             except KeyError:
-                dpi = 100
-                # Do not modify original
-                self.theme = copy(self.theme) + theme(dpi=dpi)
-
-            # Should not need this with MPL 2.0
-            fig_kwargs['dpi'] = dpi
+                self.theme = self.theme + theme(dpi=100)
+        else:
+            self.theme = self.theme + theme(dpi=dpi)
 
         if verbose:
             warn("Saving {0} x {1} {2} image.".format(


plotnine/ggplot.py
```diff
@@ -547,11 +547,12 @@ def save(self, filename=None, format=None, path=None,
                 dpi = self.theme.themeables.property('dpi')
             except KeyError:
                 dpi = 100
-                # Do not modify original
-                self.theme = copy(self.theme) + theme(dpi=dpi)
 
-            # Should not need this with MPL 2.0
-            fig_kwargs['dpi'] = dpi
```
has2k1: I think only this line needs to be moved out of the `if` statement.

yejianye: If `fig_kwargs['dpi']` is not needed with MPL 2.0, would it be using dpi in `self.theme` to determine dpi for output image?

I moved out the following line
```python
self.theme = copy(self.theme) + theme(dpi=dpi)
```
To make sure we override default dpi in the theme.

has2k1: > If fig_kwargs['dpi'] is not needed with MPL 2.0, would it be using dpi in self.theme to determine dpi for output image?

Yes

O.K I get it. That code is needlessly complicated. May be this is more clear.

```python
        # dpi
        try:
            self.theme.themeables.property('dpi')
        except KeyError:
            dpi = 100

        if dpi is not None:
            self.theme = self.theme + theme(dpi=dpi)
```

Does it make sense?

yejianye: We will only set dpi to 100 or `self.theme.property('dpi')` if dpi is not set in the options.

So you have to put 
```python
try:
            self.theme.themeables.property('dpi')
        except KeyError:
            dpi = 100
```
inside `if dpi is None` condition block.

And I think the original code is quite clear
```python
if dpi is None:
   ...

self.theme = copy(self.theme) + theme(dpi=dpi)
fig_kwargs['dpi'] = dpi```

The first part is to set default dpi if it's not set in the options. The second part is to update `theme` and `fig_kwargs` to apply dpi settings.

Comment (has2k1) [reply plotnine/ggplot.py]: Since MPL 2.0.0 is the minimum version `fig_kwargs['dpi'] = dpi` is not required. So the `theme` only needs to be changed if a `dpi` is given or is not set in the `theme`.

Comment (yejianye) [reply plotnine/ggplot.py]: Thinking of the following scenario
- MPL >= 2.0
- dpi is not given
- there's no default dpi in the theme

With your code, dpi is set to 100. But it doesn't update that dpi=100 to the theme. And fig_kwargs['dpi'] will have no effect since MPL >= 2.0.

Comment (has2k1) [reply plotnine/ggplot.py]: I see the confusion. The first `if` statement is not there any more.

```python
if dpi is None: # <- removed
    try:
        ...
```

has2k1: You are right. I'm trying avoid a duplicate setting of the theme and also be explicit about the cases involved. Maybe this won't confuse me next time I hurriedly look at it.

```python
        # dpi
        if dpi is None:
            try:
                self.theme.themeables.property('dpi')
            except KeyError:
                dpi = 100
                # Do not modify original
                self.theme = self.theme + theme(dpi=dpi)
        else:
            self.theme = self.theme + theme(dpi=dpi)
```

Also, the `copy`ing of the theme is unnecessary.

yejianye: hmm... I'm not sure whether this is 100% correct either. 

I assume you still need have 
```python
fig_kwargs['dpi'] = dpi
```
After this code block to support MPL < 2.0.

If dpi is not given and self.theme has dpi property. In that case, `fig_kwargs['dpi']` would be set to None. And I'm not sure whether that would work for MPL < 2.0.

To follow your logic, the correct code should be
```python
        # dpi
        if dpi is None:
            try:
                dpi = self.theme.themeables.property('dpi')
            except KeyError:
                dpi = 100
                self.theme = self.theme + theme(dpi=dpi)
        else:
            self.theme = self.theme + theme(dpi=dpi)
        
       fig_kwargs['dpi'] = dpi
```

If you're fine with that code, I will update the commit.

has2k1: Yeah, update the commit. The minimum supported version of [MPL is 2.0.0](https://github.com/has2k1/plotnine/blob/2b5c89f7b65a23072691fc7ce56f8b382f32d85e/setup.py#L51).

Maybe I will add a test after you do so.

Thank you for your patience on this.

yejianye:
Updated according to the review.

has2k1:
Thanks again.

Issue #45, opened 2017-08-14
Animating p9 figures using ArtistAnimation class of matplotlib


gokceneraslan:
Hi,

I'm trying to make animations using p9 figures, but I end up with matplotlib exceptions. I was wondering whether there is a convenient way to make animations out of plotnine plots.

Here is what I've tried so far:

```python
from plotnine import *
from plotnine.data import *

from matplotlib.animation import ArtistAnimation
import matplotlib.pyplot as plt
import numpy as np

def plot1(y):
    return plt.scatter(y[:, 0], y[:, 1], c='black'),

def plot2(y):
    return (qplot(y[:, 0], y[:, 1], xlab='x', ylab='y') +
            theme_minimal()).draw(),

# Use mtcars as toy data
X = mtcars[['disp', 'hp']].as_matrix()

# Add little noise to make animation cool
data = [X+np.random.normal(0, 1, (X.shape[0], X.shape[1])) for _ in range(50)]

fig = plt.figure(figsize=(8, 8))
artists = [plot1(x) for x in data]
ani = ArtistAnimation(fig, artists, interval=100, repeat_delay=500)
ani.save('/tmp/animation.mp4')

fig = plt.figure(figsize=(8, 8))
artists = [plot2(x) for x in data]
ani = ArtistAnimation(fig, artists, interval=100, repeat_delay=500)
ani.save('/tmp/animation2.mp4')
```

Here is the exception I get:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-39-98cef8b8381e> in <module>()
     25 artists = [plot2(x) for x in data]
     26 ani = ArtistAnimation(fig, artists, interval=100, repeat_delay=500)
---> 27 ani.save('/tmp/animation2.mp4')

~/.miniconda3/lib/python3.6/site-packages/matplotlib/animation.py in save(self, filename, writer, fps, dpi, codec, bitrate, extra_args, metadata, extra_anim, savefig_kwargs)
   1055                 for anim in all_anim:
   1056                     # Clear the initial frame
-> 1057                     anim._init_draw()
   1058                 for data in zip(*[a.new_saved_frame_seq()
   1059                                   for a in all_anim]):

~/.miniconda3/lib/python3.6/site-packages/matplotlib/animation.py in _init_draw(self)
   1374         # Flush the needed figures
   1375         for fig in figs:
-> 1376             fig.canvas.draw_idle()
   1377 
[...]

PR #59, opened 2017-09-22
Add PlotnineAnimation

has2k1:


has2k1:
@gokceneraslan, can you try this out.

See the [example](https://github.com/has2k1/plotnine-examples/blob/master/plotnine_examples/examples/PlotnineAnimation.ipynb)

Thanks.

has2k1:
Ping @gokceneraslan.

gokceneraslan:
I tried it with the following code:

```python
from plotnine import *
from plotnine.data import *
from plotnine.animation import PlotnineAnimation

import numpy as np

def plot(y):
    return (qplot(y[:, 0], y[:, 1], xlab='x', ylab='y') +
            theme_minimal())

# Use mtcars as toy data
X = mtcars[['disp', 'hp']].as_matrix()

# Add little noise to make animation cool
data = [X+np.random.normal(0, 1, (X.shape[0], X.shape[1])) for _ in range(50)]

figs = [plot(x) for x in data]
ani = PlotnineAnimation(figs, interval=100, repeat_delay=500)
ani.save('/tmp/animation.mp4')
```

and it looks really amazing, thanks a lot! #45 can be closed now, I guess.

has2k1:
Do you find the docstring instructions clear enough? Specifically the **Note** stuff. If not understood it is easy to create an inaccurate/wrong animation.

gokceneraslan:
I think it's clear. It'd be even better to have a mechanism to check whether scales are consistent and print a warning/error, if not.

By the way, did you by any chance test this with more "unusual" plots like the ones with geom_text, geom_tile, geom_smooth, facetting etc?

has2k1:
I did consider checking the scales, but I thought it could not be done reliably. Huh, now I think it can be done; I hope it holds up. Thanks.

> By the way, did you by any chance test this with more "unusual" plots like the ones with geom_text, geom_tile, geom_smooth, facetting etc?

facetting, yes. It is not in this PR, in the preparation for this commit there some was a (refactoring)[https://github.com/has2k1/plotnine/commit/796a7db764b395ade9f8d00b7e29579a770e3d3b] that targeted facets. For the rest, I saw no reason to bug hunt.

PR #118, opened 2018-02-25
[sketch] Fix scale_color_cmap to work on discrete cmaps

jdanbrown:
- `scale_color_cmap`/`scale_fill_cmap` currently only work on continuous cmaps, which makes them fail when naively combined with discrete values
- Fixing this seems to only require switching on whether the input `cmap` is discrete or continuous and delegating to `scale_*_manual` in the discrete case, but I'm not sure how you want to handle stuff like code org, docstrings, `@document`, and tests so I made this quick PR as a sketch
- Does this look like the right approach? What more should I do to make it able to merge?

### DIFF
--- a/plotnine/scales/scale_color.py
+++ b/plotnine/scales/scale_color.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 from warnings import warn
 
+import matplotlib as mpl
 from mizani.bounds import rescale_mid
 from mizani.palettes import (hue_pal, brewer_pal, grey_pal,
                              gradient_n_pal, cmap_pal,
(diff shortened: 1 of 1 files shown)


has2k1:
Thanks identifying an issue and trying to take a shot at it.

The convention is to have the `scale_*` be either continuous or discrete and where you have a continuous scale with a discrete counterpart, to append `_d` to the name. So we would have `scale_color_cmap_d` and `scale_fill_cmap_d`.

I also want to limit the use of Matplotlib to parts where it is required for plotting. This functionality would have to be implemented across both mizani and plotnine, that is; create a discrete palette in mizani and use it to create a discrete scale in plotnine.

If that is complicated, I can implement it.

jdanbrown:
Hmm, are you proposing to leave the existing `scale_color_cmap`/`scale_fill_cmap` as is and introduce new functions `scale_color_cmap_d`/`scale_fill_cmap_d` that do the `scale_*_manual` behavior I have in this PR? If so then I can make that change here easily, but I don't fully understand the mizani concern, so if there's more to it then I'd be happy for you to knock it out instead.

has2k1:
The organisation is; the real palettes are implemented in mizani and plotnine uses them.

I will take over.

jdanbrown:
Sgtm, thanks!

PR #122, opened 2018-03-02
Removed erroneous extra parenthesis in example code.

bevankoopman:
The example code fails with syntax error. Removed erroneous extra parenthesis.


has2k1:
Thanks!

PR #128, opened 2018-03-22
Fix failing formula without dot in facet_grid

JarnoRFB:
The facet_grid was failing if a dot was not specified for single
row/colum formulas. See test case
test_facet_grid_scales_free_y_formula_no_dot. It was failing with a
cryptic error, because somewhere down the line some code was
interpredeted as an imcomplete string. This is in contrast to R's
ggplot2, wich just seems to ignore missing dots. A dot is now just added
silently if missing. Alternative an exception could be thrown, but I
think this is more userfriendly. At least it should not fail with a
cryptic error as before. Moreover, scikit-misc is added as a dev 
dependency so that the smoothing tests do not fail locally.

### DIFF
--- a/plotnine/tests/test_facets.py
+++ b/plotnine/tests/test_facets.py
@@ -91,11 +91,12 @@ def test_facet_grid_scales_free_y():
          + theme(panel_spacing_x=0.3))
     assert p == 'facet_grid_scales_free_y'
 
+
 def test_facet_grid_scales_free_y_formula_dot_notation():
-    p = (g
-         + facet_grid('. ~ var1>2', scales='free_y')
+    p = (g+facet_grid('. ~ var1>2', scales='free_y')
          + theme(panel_spacing_x=0.3))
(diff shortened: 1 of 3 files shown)


plotnine/tests/test_facets.py
```diff
@@ -91,6 +91,17 @@ def test_facet_grid_scales_free_y():
          + theme(panel_spacing_x=0.3))
     assert p == 'facet_grid_scales_free_y'
 
+def test_facet_grid_scales_free_y_formula_dot_notation():
+    p = (g
+         + facet_grid('. ~ var1>2', scales='free_y')
```
has2k1: I do not see it, but there is a funny [character](https://travis-ci.org/has2k1/plotnine/jobs/357149958#L1129) on this line. I think you need to make sure that all the spaces are true ascii spaces.

has2k1:
Thank you.

Issue #133, opened 2018-04-24
Add helper to create a pdf document with a plot per page


khaeru:
This is a request for an enhancement. I find myself writing a lot of code like this:
```python
from matplotlib.backends.backend_pdf import PdfPages
import matplotlib.pyplot as plt
import plotnine as p9

# (data loading code)

with PdfPages('output.pdf') as pdf:
    for label, group_data in data.groupby('foo'):
        plot = p9.ggplot(group_data) \
            + p9.geom_point()
            # (further geoms and plotnine stuff)
        fig = plot.draw()
        pdf.save(fig)
        plt.close(fig)  # save memory
```
and find myself wishing to do something like:
```python
plot = p9.ggplot(data) \
    + p9.facet_page('foo') \
    + p9.geom_point()
    # (further geoms and plotnine stuff)
plot.save('output.pdf')
```

At least three problems I can imagine:
- Only the PDF backend in matplotlib supports multiple-page files, and plotnine doesn't know until told to draw/save whether the user will be using PDF. If they are not, I guess, they would need to supply a callback like the `labels` argument to some scales that would generate the (base)name of multiple files; or the `filename` argument to `save()` would need to be a format string like `output-{foo}.png`; or something like that.
- “That's an abuse of the grammar of graphics; you should find a better way to plot your data.” This is true, but I find myself doing this while figuring out *how* I want to design more carefully-constructed plots.
- “This is too similar to `plotnine.Animation`,” though I think it's fairly distinct, just as the multi-page PDF and animation features of matplotlib seem to be distinct and target different use-cases.

A very brief skim suggests the existing faceting code seems more oriented at producing layouts on a single page, but with a few words of direction on how this should work I'd be happy to take a stab at it over the upcoming months.

has2k1:
I think `facet_page()` is not the best way to add such functionality. If the objective is to have multiple plots plotted to different pages of a pdf document, then it is better to use a helper method e.g. 

```python
def draw_to_pdf_pages(iterable_of_ggplots, *args, **kwargs):
    ...


plots = (
    ggplot(df) + geom_point() + facet_wrap('a ~ b') + labs(title=label)
    for label, df in data.groupby('foo')
)

draw_to_pdf_pages(plots, filename='output.pdf')
```

Such an approach allows the pages to have more disparate plots and it avoids some the problems you listed.

khaeru:
Thanks for the response!

[...]

PR #149, opened 2018-06-11
Add `save_as_pdf_pages`

khaeru:
As discussed in #133.

- [x] I added some tests for the new code. Should I move these into `test_ggsave.py` instead of the separate file they are currently in? Also look at the Travis report of test coverage.
- [x] Document and/or test setting `height`, `width` and/or `dpi` of the individual plots/pages. This is handled by `ggplot.draw()` and then respected by `PdfPages.savefig()`.
- [x] Expand docstring with both list and generator examples of using the method, to appear in the public docs.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -21,6 +21,7 @@ Plot creation
    ~watermark.watermark
    ~layer.layer
    ~animation.PlotnineAnimation
+   ~ggplot.save_as_pdf_pages
 
 geoms
 =====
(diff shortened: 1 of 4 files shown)


has2k1:
> Or should users instead have to specifically

Let it be up to the users.

khaeru:
Okay. Should be all set to merge, then!

has2k1:
@khaeru, Thanks a lot for this.

stonebig:
a release soon ?

has2k1:
Yes

PR #201, opened 2018-09-27
Added annotation_logticks

TyberiusPrime:
This pull requests  adds 'annotation_logticks()' similar to the ggplot2 one:
https://ggplot2.tidyverse.org/reference/annotation_logticks.html

This adds log-10 'log ticks' as an annotation similar to a rug to the plot.

Looks like this:
image

Example usage:
```python
import pandas as pd
import plotnine
df = pd.DataFrame({"x": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0,
                         2, 3, 4, 5]})
p = plotnine.ggplot(df, plotnine.aes('x','x*3'))
p += plotnine.geom_point()
p += plotnine.geom_text(plotnine.aes(label='x'),
                    ha='left', nudge_x=.1)
p += plotnine.scale_y_log10()
p += plotnine.scale_x_log10()
#p += plotnine.geom_rug(sides='lb', width=2, alpha=0.5) #  to check correctness
p += plotnine.geoms.annotation_logticks(sides='lb', color='red',
                                        alpha=0.8, width=(1.2,.75, 0.4),
                                       linetype='solid')

p```


Right now this is not perfect.
For one, it's a 'fake geom' that ignores it's data.

And I've failed to detect when the axis are not log10, and
it then will plot misleading tick marks. Can't for the life of me figure out how
the geom could detect it though - coord.is_linear is always true.

So please criticize away.

### CODE REVIEW

**Comment by TyberiusPrime** [comment]
Can I get some feedback on this ?
Should we go with a separate package for annotations?

**Comment by has2k1** [comment]
> Can I get some feedback on this ?
Should we go with a separate package for annotations?

I wanted to deal with the release first then look at this. It belongs in this package, we'll just have to get the iterate on implementation.

There is no way to tell whether the scale for the aesthetic is in log10, I think that diligence left to the user perhaps with a note in the documentation.

On the implementation, it should not inherit from a `geom`. The structure should be like that of [annotation](https://github.com/has2k1/plotnine/blob/077c1cd6e1d192660b2675bd7395e2e6e505b45f/plotnine/geoms/annotate.py#L8). So you can inherit from `annotation` and create a `_annotation_geom` in the `__init__` method. Also, remember that now you do not need to know the ranges, you can use `-inf` and `+inf`.

**Comment by has2k1** [comment]
>But I do need to know the ranges to decide where to place the marks ...

You could create a private `geom` that inherits from `geom_rug`. When inspected annotations should not resemble geoms, so that there is less room for things like `issubclass`, `isinstance` and `hasattr` misleading.

**Comment by TyberiusPrime** [comment]
The original annotate() in ggplot2 is just a wrapper that let's you specify  single 'fixed-but-mapped' values for a layer. It's basically just a helper so you don't have to define a data frame and a mapping and repeat every column name twice.

And our geom/annotate/annotate class is just that, a wrapped geom with the dataframe and the mapping coming build the kwargs. 

Unlike in R, where annotate returns a layer (as does every other call to geom_*() or stat_*()),
it is it's own class, thinly wrapping the geom, providing just enough to be able to add it to the plot.

Plotnine does not expose layer-objects to the user, it only creates them when adding a geom to a plot.
The annotate() class forwards __radd__ to it's geom, which creates a layer, which get's added to the plot.

I'm afraid I fail to see the point. If annotate was just a function returning the created geom, it would work just as well. Right now it's result is functionally a geom, but does not identify as one.

Why do you think this distinction between annotation and geom is important?
At the very least, I think there's a superclass for both of them hiding in there...

**Comment by has2k1** [comment]
> I'm afraid I fail to see the point. If annotate was just a function returning the created geom, it would work  just as well. Right now it's result is functionally a geom, but does not identify as one.

I want uniformity based on the current code base. It makes it easier to maintain and to extend.
Conceptually geoms and annotations are different, it is just that some annotations are better implemented by using geoms.

**Comment by TyberiusPrime** [comment]
tl;dr:  I think geoms and annotations should have a common parent class.

# Preface 
I enjoy these discussion. Sorry for being so verbose.

## annotate()'s meaning
> I want uniformity based on the current code base. It makes it easier to maintain and to extend.
> Conceptually geoms and annotations are different, it is just that some annotations are better implemented by using geoms.

But it is neither uniform (annotate is a special case right now), nor is annotate() conceptually much different from a geom.

To quote the ggplot documentation:

"This function adds geoms to a plot, but unlike typical a geom function, the properties of the geoms are not mapped from variables of a data frame, but are instead passed in as vectors" .

It is basically case 14 of 'how does the data come in' I outlined in #207',
and could arguably be integrated by allowing 'array like' data in kwargs to geoms.

These both work
```python
g += p9.annotate('point', x=0.5,y=2, color='red')
g += p9.geom_point(x=0.4, y=2, color='blue')
````

but here only the 'annotate' one works
```python
g += p9.annotate('point', x=[0.5, 0.6],y=[2, 2.5], color='red')
g += p9.geom_point(x=[0.4, 0.8], y=[2, 4], color=['blue'])
```

## annotate() API

I also want to point out that I find the 'pass in the second part of geom_* as a string' to be a horrible unpythonic API. I get it why it's done this way in R (because geom_* is not the actual Geom* object that's necessary), but passing in a  plotnine.geom_* class feels so much cleaner... Perhaps it would at least be acceptable to pass in either, if you want to adhere to the R string api.

## Annotate and faceting

annotate() does not work with faceted data - there is no (obvious?) way to specify on which facet to plot,
nor does it plot on all facets by default (as ggplot does). 

It fails with the (very unhelpful) message ```TypeError: '<' not supported between instances of 'ggplot' and 'str'```.

'Constant-value' geoms can work with faceting by passing in a DataFrame having the faceted-column(s).
This must also be done for geom_*line, which ggplot special cases here.

## the other annotations
The current ggplot2 documentation lists the following annotations
 * geom_abline()  / geom_hline() / geom_vline()
   These are full fledged geoms, accepting mappend and unmapped data. plotnine got these covered, apart from the repeat-on-every-facet aspect of ggplot.
 * annotate() 
    is a work around for the overloaded data-passing api. Plotnine has it covered. Does not support 
    faceting.
 * annotation_custom() 
    Is deep voodoo to create nested plots or 'grobs'. Real world use case?
 * annotation_logticks()
    Is conceptually a data-less geom that could arguably also be a theme-variant of axis-ticks + minor_breaks=10.
 * annotation_map() 
   is a geom that add's a 'map' - i.e. arbitrary, polygons from 'map dataframes'.
   Barely document, but probably highly useful for that specific use case.

 * annotation_raster()
   Adds bitmap images to a plot

 * borders() 
   Seems to be annotation_map() in outline mode?

They are all (but geom_*line)  ll 'layers with alternative data sources'. gglot2 implements each of them as a Geom.

Conceptually, I think they're both the same thing (layers), just that a geom draw's it's data
from a dataframe (possibly inherited) and an annotation from somewhere else.

They should both derive from a common parent class.

has2k1:
> But it is neither uniform (annotate is a special case right now), nor is annotate() conceptually much different from a geom.

I view annotations as different from geoms. Annotations are the guides that help with the interpretation of  the plotted data, in this sense plot title and subtitle constitute annotations. There some annotations drawn inside the plot panel, for these it may be convenient to draw them using geoms. Because of what annotations constitute as a whole, I prefer not to have them subclass `geom`, but rather create a `geom` member if they need to.

So the uniformity I referred to is that of the `annotate` class as implemented to be the template for other annotations. Especially those that draw within the plot panel.

> I also want to point out that I find the 'pass in the second part of geom_* as a string' to be a horrible unpythonic API.

Makes sense, the `geom` itself should be an option.

>  ... nor does it plot on all facets by default (as ggplot does).

I think that was a bug in a past release.

```python
from plotnine import *
from plotnine.data import *

(ggplot(mtcars, aes('wt', 'mpg'))
 + geom_point()
 + geom_vline(xintercept=3, color='blue')
 + annotate('vline', xintercept=4, color='red')
 + facet_wrap('am')
).save('annotations-and-facetting.png')
```

annotations-and-facetting

TyberiusPrime:
Ok, I' have submitted a seperate pull request for the annotate(geoms.geom_point...) patch, but github has some issues this morning.

For this pull request, annotation_logticks is now descendant of annotate, complete with it's own private-geom for the drawing subclassing geom_rug. Test cases have been added.

I've taken the opportunity offered by subclassing annotate to check whether the plot has log10 axis
on the necessary sides and add them if they were missing or had a different transformation.
Arguably, we should instead issue a warning, or perhaps add them only if no other scale has been
set so far. Your call, hask2k1.

TyberiusPrime:
I've added coord_flip support and altered the geom_rug doc string to reflect that the sides are after the flip.

TyberiusPrime:
I'm not convinced the auto-log10 scale is going to work out - it creates issues when a coord_flip is in play.

has2k1:
I was waiting on this one as I still have not had a chance to review the algorithm. But for the surrounding choices regarding the implementation:

1. There no need for "checking plot sanity" and the user should be able to put logticks even when the scale is not log, whether it makes sense or not.

> Some of the combinations of graphs and statistical methods may be degenerate or bizarre, but there is no moral reason to restrict them.

*The Grammar of Graphics* (Chapter 7), Leland Wilkinson

2. The `coord_flip` issue should be dealt with elsewhere, at the source of the problem. I'll make a commit to address it after I mull over it a little more.

TyberiusPrime:
1. Which is why this is a warning and not an exception. 

Full expressiveness is good, but so is warning the user when he's about to shoot himself in the foot.
Especially if the resulting wrongness may be subtle (I have seen a published plot doing log ticks wrong very recently...).

The failure mode for limits in the hundreds is ok, 
nobody's going to miss that.
image

But this is more subtle (same data / 100)
image

(this is the same failure mode that ggplot2 shows, and it doesn't warn either,
which I consider neglect.)


2. Yes please! Should I post a bug report to ggplot2? You're way more involved with that project...

TyberiusPrime:
I've refactored the tick calculation algorithm.

has2k1:
You are right about the 2nd plot being wrong yet not easy to spot. So the issue is that the solution involves too much supporting code.

I know the root cause of the coord_flip issue, I'll file a bug report.

TyberiusPrime:
Ok, I have rebased on the current master and rewritten the code to take advantage of the panel_params['scales'] (which incidentally is useful for another annotating geom I have in mind :) ).

Now I need to learn more git so I can clean up this PullRequest's commits - though simply starting with a clean branch would be easier^^.

plotnine/geoms/geom_rug.py
```diff
@@ -52,8 +52,8 @@ def draw_group(data, panel_params, coord, ax, **params):
         rugs = []
         xmin, xmax = panel_params['x_range']
         ymin, ymax = panel_params['y_range']
-        xheight = (xmax-xmin)*0.03
-        yheight = (ymax-ymin)*0.03
+        xheight = (xmax-xmin)*0.03 * params['width']
+        yheight = (ymax-ymin)*0.03 * params['width']
```
has2k1: `0.03` should be removed. See comment above.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,138 @@
...
+                          " but x-axis is not log10 - nonsensical marks show.",
+                          PlotnineWarning)
+        if (not y_is_log_10 and ('l' in sides or 'r' in sides)):
+            warnings.warn("annotation_logticks for y-axis,"
+                          " but y-axis is not log10 - nonsensical marks show.",
+                          PlotnineWarning)
+        return result
+
+    @classmethod
+    def draw_group(cls, ignored_data, panel_params, coord, ax, **params):
```
has2k1: Use `draw_panel`.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,138 @@
...
+        ([0..1, ...]])
+        """
+        result = []
+        for l in range(lower_log, upper_log + 1):
+            for m in marks_at:
+                result.append(10**(l + np.log10(m)))
+        return result
+
+    @classmethod
+    def draw_one_axis(cls, axis, min_value, max_value, panel_params, coord, ax,
```
has2k1: Rename this to `draw_group`.

TyberiusPrime:
I've implemented the requested changes.

Base other than 10 does not seem useful - the reader has to understand the grammar of the plot,
and I think anything but base 2 (no log marks), base 10, and perhaps base e are rather rare.

has2k1:
`mizani` allows for log transformations (and therefore scales) of any base, and doing the same here would not add that much code nor make it considerably less readable.

TyberiusPrime:
Sure, adding base as a parameter is easily done (log bases are all multiplicative of each other)

But where do you place the ticks on arbitrary bases?

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+            data = pd.DataFrame({
+                axis: positions,
+                'size': params['size'],
+                'color': params['color'],
+                'alpha': params['alpha'],
+                'linetype': params['linetype'],
+            })
+            super().draw_group(
+                data, panel_params, coord, ax, length=length, **params)
+
```
has2k1: The functionality in this method can be moved to `draw_panel`.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+        upper_log = int(math.ceil(max_value)) + 1
+        major = np.log10(cls._calc_ticks(lower_log, upper_log, [1.0]))
+        middle = np.log10(cls._calc_ticks(lower_log, upper_log, [0.5]))
+        minor = np.log10(cls._calc_ticks(lower_log, upper_log,
+                         [0.2, 0.3, 0.4, 0.6, 0.7, 0.8, 0.9]))
+
+        lengths = params.get('length', cls.DEFAULT_PARAMS['length'])
+        if len(lengths) != 3:
+            raise ValueError(
+                "length for annotation_logticks must be a tuple of 3 floats")
```
has2k1: The length checking can be moved to `annotation_logticks.__init__`

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+        these are the sides *after* the flip.
+
+    length: tuple (default (0.036, 0.0225, 0.012))
+        length of the ticks drawn for full / half / tenth
+        ticks relative to panel size
+    """
+
+    def __init__(self,
+                 sides='bl',
+                 **kwargs):
```
has2k1: It should have all the parameters as the `annotation` is part of the face API.

```python
    def __init__(self, sides='bl', alpha=1, color='black', size=0.5,
                 linetype='solid', lengths=(1.2, 0.75, 0.4)):

```

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+                'size': params['size'],
+                'color': params['color'],
+                'alpha': params['alpha'],
+                'linetype': params['linetype'],
+            })
+            super().draw_group(
+                data, panel_params, coord, ax, length=length, **params)
+
+
+@document
```
has2k1: The `@document` decorator is for `geoms`, `stats` and `scales`.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+    def draw_panel(cls, ignored_data, panel_params, coord, ax, **params):
+        if 'l' in params['sides'] or 'r' in params['sides']:
+            ymin, ymax = panel_params['y_range']
+            cls.draw_group('y', ymin, ymax, panel_params, coord, ax, params)
+        if 'b' in params['sides'] or 't' in params['sides']:
+            xmin, xmax = panel_params['x_range']
+            cls.draw_group('x', xmin, xmax, panel_params, coord, ax, params)
+
+    @staticmethod
+    def _calc_ticks(lower_log, upper_log, marks_at):
```
has2k1: It is better of `_calc_ticks` just take a range. Maybe something like:

```python
@staticmethod
def _calc_ticks(range):
    
    def _calc(positions):
        ...

    major = _calc([1.0])
    middle = _calc([0.5])
    minor = _calc([.1, 0.2, 0.3, 0.4, 0.6, 0.7, 0.8, 0.9])
    return major, middle, minor
```

Where the `(major, middle, minor)` tick positions are already log transformed. This has all the calculations happen in one appropriately named function.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+                          " but x-axis is not log10 - nonsensical marks show.",
+                          PlotnineWarning)
+        if (not y_is_log_10 and ('l' in sides or 'r' in sides)):
+            warnings.warn("annotation_logticks for y-axis,"
+                          " but y-axis is not log10 - nonsensical marks show.",
+                          PlotnineWarning)
+        return result
+
+    @classmethod
+    def draw_panel(cls, ignored_data, panel_params, coord, ax, **params):
```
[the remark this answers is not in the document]
has2k1: Let me know if you have difficulty moving the functionality into this method.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
+import math
```
has2k1: I think `numpy` is enough. You can do `np.floor` and `np.ceil`.

plotnine/geoms/annotation_logticks.py
```diff
@@ -0,0 +1,139 @@
...
+        'alpha': 1,
+        'color': 'black',
+        'size': 0.5,
+        'linetype': 'solid',
+        'length': (0.036, 0.0225, 0.012),
+    }
+    legend_geom = 'path'
+
+    @staticmethod
+    def _check_log_scale(sides, scales, is_coord_flip):
```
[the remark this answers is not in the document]
TyberiusPrime: It's not unnecessary, it's loose coupling, which makes testing easier (no need for mock objects).

But I have not taken it far enough, it should not take either scaler or coord, but only tthat which it depends on.

plotnine/geoms/annotation_logticks.py
TyberiusPrime: It could, I'd rather not .
It would require a temporary list of tuples and a loop and an unintuitive control flow compared to two ifs+function calls (remember, you can have them on both axis in the same plot).

plotnine/geoms/annotation_logticks.py
TyberiusPrime: I've refactored accordingly, but the inner _calc is till a class level (static) function so it can be tested.

plotnine/geoms/annotation_logticks.py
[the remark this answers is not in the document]
TyberiusPrime: see below.

has2k1:
Thanks for pushing through on the reviews, as the tests pass can you rebase and squash into one commit with message `Add annotation_logticks`, and also a not in `doc/changelog.rst` in the *New Features* section. I think I can make the other changes a lot faster than I can share my mind-frame.

TyberiusPrime:
Done. Looking forward to see what you make of it.

has2k1:
This PR has brought about a change that need to happen in mizani. `minor_breaks` for log scales should not be set to a fixed number. They should be two less than the integer base, hence for base 10 there is no need to have `minor_breaks=8`.

Issue #218, opened 2018-10-25
Add adjustText (for text repelling) to geom_text


has2k1:
[adjustText](https://github.com/Phlya/adjustText).

See also:
1. https://gist.github.com/votti/885672c90b4550154215e07190730a1b
2. https://github.com/Phlya/adjustText/wiki

TyberiusPrime:
That's cool.

It think ggrepel does it as a custom geom - I guess it would be hard to do it as a position, 
since that happens way earlier?

TyberiusPrime:
See pull request #220

PR #220, opened 2018-10-30
introduced position_adjust_text

TyberiusPrime:
Here's a first implementation of adust_text using a position - see #218 
I have extended the position to allow a post-drawing callback,
and positions can now define a list of geoms that they work with.

While this enables the basic use case, there are some kinks we should probably talk with upstream about:

```python
(p9.ggplot(mtcars, p9.aes('mpg','disp')) + p9.geom_point()
      + p9.geom_text(p9.aes(label='name'), color='blue', position = p9.position_adjust_text(), size=7)
    )
```
image

1) Two sets of points and labels don't work well:
```python
g = (p9.ggplot(mtcars, p9.aes('mpg','disp'))     
     + p9.geom_point()     
      + p9.geom_text(p9.aes(label='name'), color='blue',
                      position = p9.position_adjust_text({
          'arrowprops': {'arrowstyle': '-', 'color': 'blue'}
      }), size=7)
     + p9.geom_point(p9.aes('mpg+1','disp'), color='green')
     + p9.geom_text(p9.aes(x='mpg+1', label='name'), color='green',
                      position = p9.position_adjust_text({
          'arrowprops': {'arrowstyle': '-', 'color': 'green'}
      }), size=7)
    )
g
```
image
(I think it's an issue with both set's of text labels being drawn on the same ax, and accordingly
the second call sees the first set again. Will have to debug.)

2) labels and arrows interact badly
```python
(p9.ggplot(mtcars, p9.aes('mpg','disp')) + p9.geom_point()     
      + p9.geom_label(p9.aes(label='name'), color='blue', position = p9.position_adjust_text(), size=7)
    )
```
image

3) no geom_point() -> broken plot
```python
(p9.ggplot(mtcars, p9.aes('mpg','disp')) 
     # + p9.geom_point()     
      + p9.geom_text(p9.aes(label='name'), color='blue', position = p9.position_adjust_text(), size=7)
    )
```
image

3b) geom_point() after geom_text(position_adjust_text) > broken plot
```python
(p9.ggplot(mtcars, p9.aes('mpg','disp'))      
      + p9.geom_text(p9.aes(label='name'), color='blue', position = p9.position_adjust_text(), size=7)
      + p9.geom_point()     
    )
```
image

Also, I note that the adjustment is not fast - it takes a few seconds ever for these 'simple' plots.


has2k1:
This is promising. I hoped to implement it without changing the `geom` API or creating a new `position`. `geom_text` already accommodates for extra adjustment using the `nudge_*` parameters, I planned to introduce another parameter for `adjust_text`.

TyberiusPrime:
That's fairly doable, though I will point out that using nudge_* will silently replace the text's position with a position_nudge...

So we'd have geom_text(adjust_text = True or dict_of_adust_text_parameters), right?
Would throw a PlotnineError if used together with nudge_*, and would trigger
a adjust_text call in geom_text.draw_panel()... (same in geom_label).

has2k1:
> I will point out that using nudge_* will silently replace the text's position with a position_nudge...

Yes, nudge_* are just for convenience given how common it is that nudging is needed with texts.
Similarly, `adjust_text={}` should be enough for default repelling if the package is installed.

This is the docstring I had for the implementation.

```python
"""
    adjust_text: dict (default: None)
        Parameters to :class:`adjustText.adjust_text` will repel
        overlapping texts. This parameter takes priority of over
        ``nudge_x`` and ``nudge_y``.
"""
```

TyberiusPrime:
Ok, I've implemented it as a parameter. Much less work, actually. 

Test cases are separate for geom_label and geom_text, in case their implementation drifts apart later on.

TyberiusPrime:
should I add adjustText to the requirements or do we keep it optional?

has2k1:
> should I add adjustText to the requirements or do we keep it optional?

It should be optional, should be installed for the tests and documentation.

Add it to these places

1. [conf.py](https://github.com/has2k1/plotnine/blob/4cccc40a2388ad29783385cedf50e273851b673a/doc/conf.py#L412)
2. [travis.yml](https://github.com/has2k1/plotnine/blob/4cccc40a2388ad29783385cedf50e273851b673a/.travis.yml#L60)
3. [requirements/rtd.txt](https://github.com/has2k1/plotnine/blob/4cccc40a2388ad29783385cedf50e273851b673a/requirements/rtd.txt)
4. [requirements/dev.txt](https://github.com/has2k1/plotnine/blob/4cccc40a2388ad29783385cedf50e273851b673a/requirements/dev.txt)

plotnine/tests/test_geom_text_label.py
```diff
@@ -49,3 +49,37 @@ def test_label_aesthetics():
...
+            mtcars.tail(2),  # two points -> fast test.
+            aes('mpg', 'disp', label='name')) + geom_point(
+                size=5, fill='black') + geom_text(
+                        adjust_text={
+                            'expand_points': (2, 2),
+                            'arrowprops': {
+                                'arrowstyle': '->',
+                                'color': 'red'
+                            }
+                        }))
```
has2k1: Can you change the styling for the test to something like

```python
    adjust_text={
        'expand_points': (2, 2),
        'arrowprops': {
            'arrowstyle': '->',
            'color': 'red'
        }
    }

    p = (ggplot(mtcars.tail(2), aes('mpg', 'disp', label='name'))
         + geom_point(size=5, fill='black')
         + geom_text(adjust_text=adjust_text)
         )
```

That way, it is easy to tell which layers are in the plot and also to comment out any layer. In some of the earlier tests the `+` signs are on the right (not to my exact preference) since the pep8 standard (and the linting tools) was broken, but it has since been [fixed](https://github.com/python/peps/commit/7690a00d478866eb3cae3d82bbe1e80199a7392f#diff-64ec08cc46db7540f18f2af46037f599).

plotnine/tests/test_geom_text_label.py
```diff
@@ -49,3 +49,37 @@ def test_label_aesthetics():
          scale_y_continuous(limits=(-0.5, n-0.5)))
 
     assert p == 'label_aesthetics'
+
+
+def test_adjust_text():
+    from plotnine.data import mtcars
```
has2k1: Put it at top of file.

TyberiusPrime: I presume you mean the import and not the test.

has2k1:
> That's necessary to stop flake8 from complaining though.

Really! What version of flake8?

TyberiusPrime:
The one run by travis? At any rate, there shouldn't be a an empty line at the end of test_postion.py...  I had removed that in the flake8 commit, now it's reintroduced and the flake8 check fails again.

I'm not sure where I dragged it in though, since the current master doesn't seem to have it.

has2k1:
Thanks for the iterations, I made some changes and merged.

TyberiusPrime:
Awesome, thank you.

You have a consistently mispelled variable in geom_text.py: HAS_ADJSUST_TEXT

has2k1:
> You have a consistently mispelled variable in geom_text.py: HAS_ADJSUST_TEXT

Thanks for catching the slip up.

has2k1:
> Can't find documentation/examples how to use positiion_adjuts_text. Is this feature working?

It is has not yet been released.

TyberiusPrime:
If you're using the github master though, you can try it like this:
```python
plot += geom_text(aes(...), adjust_text = {
    'expand_points': (2, 2),
    'arrowprops': {
        'arrowstyle': '->',
        'color': 'red'
    }
})
```

gballardin:
Was this feature ever implemented? Not sure how nuanced it is to add it to the core API, but it would be a very useful addition.

Issue #225, opened 2018-11-02
edge color for labels


TyberiusPrime:
I'd like to introduce edge colors for labels.

I can either expose them as (unmapped) parameter, or as a full aesthetic,
but I run into the issue that I then need a third color scale, besides color and fill.

What's the preferred way to tackle this?

has2k1:
Use a parameter for `box_color` so that if specified it overrides the `color`.

TyberiusPrime:
done in #226

PR #226, opened 2018-11-02
added boxcolor to geom_label

TyberiusPrime:
Improvement closing #225


plotnine/geoms/geom_label.py
```diff
@@ -15,6 +15,7 @@
     #
     # Documentation at matplotlib.patches.BoxStyle
     'boxstyle': 'round',
+    'boxcolor': 'black',
```
has2k1: Set the default to `None` so that `boxcolor` only has an effect if the user has set it.

has2k1:
Avoid my `dev` branch. I rebase often.

plotnine/geoms/geom_label.py
TyberiusPrime: I've restored the old behaviour: unset boxcolor means 'use color aes' and documented accordingly.

has2k1:
Thanks. I did some amending, they are now in master.

PR #230, opened 2018-11-13
introduced annotation_alternating_backgrounds

TyberiusPrime:
This PR introduces an annotation that adds 'alternating'  backgrounds,
which is useful to increase readability in tight jitter plots on categorical x-axis data.
Example:
```python
import plotnine as p9
g = p9.ggplot(mtcars.assign(
    gear = pd.Categorical(mtcars.gear),
    am = pd.Categorical(mtcars.am)
))
g += p9.geom_jitter(p9.aes("gear", 'wt', color='am'), width=.45, height=0)
g += p9.annotation_alternating_background(fills=['grey','darkgrey'],                                         
                                          alpha=.5)
g += p9.scale_color_manual(['red','blue'])
```
image

Supports coord_flip:
image

And more than 2 colors, and does something sensible with non-discrete axis 
(ie. change shade halfway between labels)
image


plotnine/geoms/annotation_alternating_background.py
```diff
@@ -0,0 +1,127 @@
...
+            **kwargs)
+
+    def __radd__(self, gg, inplace=False):
+        # background is always the lowest layer
+        l = layer.from_geom(self._annotation_geom)
+        l._insert(0, gg)
+        return gg
+
+
+class _geom_annotation_alternating_background(geom):
```
has2k1: `_geom_alternating_background`

plotnine/geoms/annotation_alternating_background.py
```diff
@@ -0,0 +1,127 @@
...
+    Default: ["#AAAAAA", "#CCCCCC"]
+
+    {common_parameters}
+    """
+
+    def __init__(self, **kwargs):
+        self._annotation_geom = _geom_annotation_alternating_background(
+            **kwargs)
+
+    def __radd__(self, gg, inplace=False):
```
has2k1: There is no need for this __radd__.

Being extra helpful to the user by forcing the this annotation to be the lowest layer has two consequences.

1. It limits how creative the user can be with the layer.
2. The code in other places has to be broken-up and rearranged to suit this functionality.

The burden should be on the user to put the layer created by this annotation in the right position. Users of this library learn quickly that they have to manage the order of the layers.

plotnine/geoms/annotation_alternating_background.py
```diff
@@ -0,0 +1,127 @@
...
+    REQUIRED_AES = set()
+    DEFAULT_PARAMS = {
+        "stat": "identity",
+        "position": "identity",
+        "na_rm": False,
+        "color": None,
+        "fills": ["#AAAAAA", "#CCCCCC"],
+        "linetype": "solid",
+        "size": 0,
+        "alpha": 0.5,
```
has2k1: The `direction` should be a parameter e.g `'direction':  'vertical'`

plotnine/geoms/annotation_alternating_background.py
```diff
@@ -0,0 +1,127 @@
...
+    }
+    legend_geom = "polygon"
+
+    @staticmethod
+    def draw_group(data, panel_params, coord, ax, **params):
```
has2k1: On the implementation, this `geom` should create the right data and leave the rest to `geom_rect`.

plotnine/layer.py
```diff
@@ -200,7 +200,20 @@ def __radd__(self, gg):
...
+        Insert a layer into a ggplot object
+        """
+        try:
+            gg.layers.insert(position, self)
+        except AttributeError:
+            msg = "Cannot add layer to object of type {!r}".format
+            raise PlotnineError(msg(type(gg)))
+        return self.__update_from_layer(gg)
+
+    def __update_from_layer(self, gg):
```
has2k1: As noted above, the changes in this file are not necessary.

TyberiusPrime:
While I'm all for expressiveness, I don't think " make me a plot with this data, and the first and most important thing about it is that it has alternating backgrounds" is what the usage of this annotation should look like.

The user should be able to add this after she has defined all the (important) data geoms.

How about we use the substraction operator to mean "insert as bottom.most from"? That would open up some flexibility in the ordering...

has2k1:
> The user should be able to add this after she has defined all the (important) data geoms.

It is not a big inconvenience to the user to position the cursor a little higher and add the very first layer. We the developers pay a penalty when we split up the code and create functions like `layer.__update_from_layer`. The code is more complicated, and there is a big special case -- "some layers are more equal than others".

> How about we use the substraction operator to mean "insert as bottom most geom"?

I think that complicates the use of the library. e.g.

```python
(ggplot()
 + layer_a
 + layer_b
 - layer_c
 + layer_d
 - layer_e
)
```

Should `layer_e` really be below `layer_c` in the plot? It takes a few second to think about it, and it is not a stretch to convince yourself of the reverse ordering.

Most users (me included) are incapable of writing plotting code left to right, top to bottom (except for the most basic plots); we are always re-positioning the cursor i.e. developing the plot.

TyberiusPrime:
> we split up the code and create functions like `layer.__update_from_layer`
That is a very minor refactoring to not copy paste the relevant code from add. It was needed in one place, now there are two places that need it, simple as that. I could have have add(...) forward to insert(len(layers), ...) instead.

It is not about moving the cursor, but about reading the plot 6 months later. All my plots
go like this: data, aes, geoms, scales * coords, annotations, theming, (and save).

Now I suddenly have to place an annotation before the geoms.

You have convinced me that having the annotation going to the bottom 'magically' isn't good,
Having to place it out of context doesn't sit well with me either though.

Maybe an operator is not explicit enough, though having a method on the ggplot object will break 
the flow of +.

How about a "function" so that your example would look like

```python
(ggplot()
 + layer_a
 + layer_b
 + place_at_bottom(layer_c)
 + layer_d
  + place_at_bottom(layer_e)
)
```
Or perhaps place_at_z(0, layer_c) and then the user is free to place e either above or below c with 
place_at_z(0, ...) or place_at_z(1, ...).


--
Either way it would be nice to expose an insert(position, layer) from the plot object,
then [dppd-plotnine](https://pypi.org/project/dppd-plotnine/) could offer place_at_bottom without duplicating the magic from _radd_ or fiddeling with the internal layer list.


--
May you have a happy new year!.

has2k1:
> It is not about moving the cursor, but about reading the plot 6 months later.

I organise and comment the more complicated plots. I do the same for the plots in the gallery for [example](https://plotnine.readthedocs.io/en/stable/generated/plotnine.geoms.geom_segment.html#examples), so I tend to have minimal readability issues later on.

> How about a "function" so that your example would look like

I would not rush to have such a function in the library, my first opinion is that it would be in bad taste. But users should be able to create one for their toolbox and if it turns out to be a good idea, then I can add one in plotnine. Other packages should also be free to experiment with how the layers are ordered.

The `layer.__radd__` function does not have a clear separation of duties. I think I can refactor the code a little and make it easier for your package(s) to insert layers. 

---
I see you have also taken a crack at dplyr. Nice, keep it up, I hope it matures well.

has2k1:
`annotation_alternating_background` is a mouthful, what about `annotation_stripes`?

TyberiusPrime:
That sounds most reasonable @has2k1 with regards to the insert-functionality.

I'll trim this pull request down till it's just annotation_stripes() (I like that name).
As for the direction, do we go with 'vertical/horizontal' or 'x'/'y', or support both?
I don't think we need direction: 'both', that can be covered by adding two annotation layers
with alpha transparency.

has2k1:
Use 'vertical' and 'horizontal'.

> I don't think we need direction: 'both', that can be covered by adding two annotation layers
with alpha transparency.

Yes, 'both' is problematic because the order can matter.

Also can you add an `extent` parameter with default values of `(0, 1)`. This would control the region occupied by the stripes. Such that `(0, 1)` means full length, `(0, 0.5)` first half, `(0.25, 0.75)` middle half, ....

TyberiusPrime:
All right, after sleeping a night on it I've convinced myself that you meant that the extend is orthogonal to the stripes.

I've implemented it as such, and the requested changes have been done as well.

has2k1:
The `extend` parameter would make it possible to have stripes that do not cover the full range of the axis. e.g `extend=(0.1, 0.9)`.

It would also be helpful to add a `width` parameter.

I can add them after you.

> What is the use case?

It makes creating row or column patterns easier and there is more room for creativity, as you would not need to generate a dataframe to create the desired effect, [see](https://plotnine.readthedocs.io/en/stable/generated/plotnine.geoms.geom_segment.html#ranges-of-similar-variables), the horizontal stripes are the same color but not full full extent, so there is room for other information, and they are not full `width` hence they appear as stripes instead of just one background color.

TyberiusPrime:
Ok, then I was correct in that extend for 'vertical' stripes describes what they cover on the vertical axis, like this:
image
extend = (0.3, 0.75).

Would width shrink toward the center or toward the left/bottom? 
Please note that currently the stripes are not equal sized - the first and last one are larger to cover the 'expand' of the scale.

TyberiusPrime:
I'm unclear on what we need to move this forward, please advise.

has2k1:
> I'm unclear on what we need to move this forward, please advise.

I will finish this up, I already have a few changes from a while back.

has2k1:
@TyberiusPrime, thanks a lot on this. I have made a few changes after merging.

TyberiusPrime:
Thank you!

PR #232, opened 2018-11-23
input validation on geom_violin(draw_quantiles)

TyberiusPrime:
This PR checks, normalizes and documents the draw_quantiles parameter on geom_violin.

It now accepts any iterable of floats, or a single float (to be feature compatible with ggplot2).
They get converted into a list internally.

If it's not valid (ie. not float, [float] or values outside of 0..1 (exclusive), a ValueError is raised
when calling geom_violin.

Previously, the following happened:
 * draw_quantiles = True -> Exception in drawing  ```TypeError: object of type 'bool' has no len()``` 
 * draw_quantiles = 0.5 -> Exception in drawing: ```TypeError: object of type 'float' has no len()```
 * draw_quantiles = np.array([0.5]) -> ok
 * draw_quantiles = np.array([0.3, 0.5]) -> exception in drawing: ```ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()```
* draw_quantiles = 0 -> interpreted as False
* draw_quantiles = [0.0] -> exception in drawing: ```ValueError: A value in x_new is below the interpolation range.```
* draw_quantiles = [1.0] -> exception in drawing: ```ValueError: A value in x_new is above the interpolation range.```
* draw_quantiles = [1.25] -> exception in drawing: ```ValueError: A value in x_new is above the interpolation range.```

These all now either work, or throw a sensible exception ```ValueError("draw_quantiles must be a float or an iterable of floats (>0.0; < 1.0)")``` much closer to the actual erroneous call.

### DIFF
--- a/plotnine/geoms/geom_violin.py
+++ b/plotnine/geoms/geom_violin.py
@@ -19,6 +19,9 @@ class geom_violin(geom):
     Parameters
     ----------
     {common_parameters}
+    draw_quantiles: float or [float]
+       draw horizontal lines at the given quantiles (0..1)
+       of the density estimate.
     """
     DEFAULT_AES = {'alpha': 1, 'color': '#333333', 'fill': 'white',
                    'linetype': 'solid', 'size': 0.5, 'weight': 1}
@@ -27,6 +30,17 @@ class geom_violin(geom):
                       'draw_quantiles': None, 'scale': 'area',
                       'trim': True, 'width': None, 'na_rm': False}
     legend_geom = 'polygon'
+
+    def __init__(self, *args, **kwargs):
+        if 'draw_quantiles' in kwargs:
+            kwargs['draw_quantiles'] = pd.Series(kwargs['draw_quantiles'],
+                                                 dtype=float)
+            if not kwargs['draw_quantiles'].between(0, 1,
+                                                    inclusive=False).all():
+                raise ValueError(
+                    "draw_quantiles must be a float or"
+                    " an iterable of floats (>0.0; < 1.0)")
+        super().__init__(*args, **kwargs)
 
     def setup_data(self, data):
         if 'width' not in data:
@@ -76,7 +90,7 @@ class geom_violin(geom):
             geom_polygon.draw_group(polygon_df, panel_params,
                                     coord, ax, **params)
 
-            if quantiles:
+            if quantiles is not None:
                 # Get dataframe with quantile segments and that
                 # with aesthetics then put them together
                 # Each quantile segment is defined by 2 points and
--- a/plotnine/tests/test_geom_violin.py
+++ b/plotnine/tests/test_geom_violin.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+import pytest
 
 from plotnine import ggplot, aes, geom_violin, coord_flip, theme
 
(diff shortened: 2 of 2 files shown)


TyberiusPrime:
Bumping this with a request for feedback :).

has2k1:
It is okay to do validation and give the user useful feedback, but since python dynamic, if we do this more places it would add a lot of code and impact the readability. However, I have want to improve the usability and stability of plotnine. To this end, I have thought about using a validation library which impose a structure on how we do validations. The main candidate is [traitlets](https://github.com/ipython/traitlets), which would be a huge undertaking and potentially other consequences that I have not considered.

For this particular case, I think a different implementation would be better.

1. Do the validation in `setup_params` method.
2. Use `utils.make_iterable` function
3. Add another function `in_range`

    ```python
    def in_range(x, range=(0, 1), include=(True, True)):
        ...
    ```

    The best place for this function is in `mizani.bounds`.

Then we have something like

```python
params['draw_quantiles'] = make_iterable(params['draw_quantiles'])
if not in_range(params['draw_quantiles'], (0, 1), include=(False, False)):
    raise ValueError("...")
```

TyberiusPrime:
1. No other geom has a setup_params, suppose you meant setup_data.

That's a suboptimal place though, because setup_data is called when the plot is being build (drawn), but we can fail much earlier - the test in the __init__. 
That's much nicer for the user since the traceback pinpoints the wrong call.

2. Sure, I've done so.

3. Python has a perfectly good 'in range' comparison (lower < x < upper). 
A function in_range would only complicate matters and is in no way clearer.

has2k1:
> No other geom has a setup_params, suppose you meant setup_data.

Oh yes. `setup_params` is only available for `stats`.

> Python has a perfectly good 'in range' comparison (lower < x < upper)

I have a vectorized function that is part of my tools and would do for this purpose and elsewhere.
The range comparison in python is one scalar value at a time!

```python
def in_range(x, range=(0, 1), include=(False, False)):
    """
    Return True if all values in x are in the given range

    Parameters
    ----------
    x : array-like
        Values to be checked
    range : tuple
        (min, max) values of the range
    include : tuple
        Whether the min and max values are included in
        the range

    Returns
    -------
    out : bool
        Whether all values in x are in the range.
    """
    x = np.asarray(x)
    greater = np.greater_equal if include[0] else np.greater
    less = np.less_equal if include[1] else np.less
    return np.all(greater(x, range[0]) & less(x, range[1]))
```

TyberiusPrime:
Fair point on the vectorization.
But you still need a type check, since all you have is an iterable, not an array.
Also we are.missing Nan checking right now.

Personally I'd call that function all_in_range since that what it checks.

has2k1:
> But you still need a type check, since all you have is an iterable, not an array.

That is an edge case to be avoided by modifying `make_iterable` to meet what it is intended for. i.e You get a list-like out. So if an `iter` object goes in, a list comes out.

TyberiusPrime:
> > But you still need a type check, since all you have is an iterable, not an array.
> 
> That is an edge case to be avoided by modifying `make_iterable` to meet what it is intended for. i.e You get a list-like out. So if an `iter` object goes in, a list comes out.

I meant when you receive non floats.

Anyway, pandas.Series.between().all() does exactly what we need without the need for us to introduce in_range - I've changed the PR accordingly.

has2k1:
That is okay. Can you change the commit message to `Validate draw_quantiles parameter in geom_violin
`

TyberiusPrime:
Sure.

TyberiusPrime:
Something upstream must have changed - these tests pass locally, and my change shouldn't be affecting those tests at all.

has2k1:
Matplotlib v3.1.0 shipped. I'll look into the issue.

TyberiusPrime:
I can confirm that upgrading to the newest matplotlib also triggers a lot of failing tests locally.

Seems to be something with the ticks.

TyberiusPrime:
Any news on this?
Even if we don't get compability with the newest matplotlib right away, it might be possible to verify this with the older one for now?

This PR is getting old enough to walk...

has2k1:
> Even if we don't get compability with the newest matplotlib right away, it might be possible to verify this with the older one for now?

The tick issue at the heart of the failing tests requires big changes to how theming is applied. I want to fix that first even if it is a temporarily solution. After that this PR should work with rebase.

has2k1:
@TyberiusPrime, please rebase. I have fixed the failing tests.

TyberiusPrime:
I have rebased - sorry for the delay.

has2k1:
@TyberiusPrime, thank you!

PR #243, opened 2019-01-05
Refactor the creating of layers

has2k1:
Separate the addition of objects (geoms, stats and annotations)
to the main ggplot plot object from the creation of labels for
these objects.

Add `to_layer()` to the internal API of all objects that create
layers.


has2k1:
@TyberiusPrime, this should allow you to manipulate layers more freely. See the [test](https://github.com/has2k1/plotnine/pull/243/files#diff-e1b33f558b74cef6f0daa8c54b4e891eR62) case.

TyberiusPrime:
Awesome

PR #247, opened 2019-01-20
Add library imports to documentation homepage

kingishb:
👋  As a new user, I tried to run the code example to generate the provided `mtcars` plot. However, I didn't know what to import to get all the functions loaded and the `mtcars` dataset. Adding

```
from plotnine import *
```

can help new users try out the library without seeking any further docs or examples.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -19,7 +19,9 @@ simple plots remain simple.
 Example
 -------
 .. code:: python
-
+    from plotnine import *
+    from plotnine.data import mtcars
+    
     (ggplot(mtcars, aes('wt', 'mpg', color='factor(gear)'))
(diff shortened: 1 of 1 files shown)


doc/index.rst
```diff
@@ -19,7 +19,9 @@ simple plots remain simple.
 Example
 -------
 .. code:: python
-
+    from plotnine import *
```
has2k1: Can you use an explicit import

```python
from plotnine import ggplot, geom_point, aes, stat_smooth
```

kingishb: Sure thing, just changed.

has2k1:
@kingishb, Thanks.

PR #250, opened 2019-01-22
Semi-automatic limits in transformed axis, Breaks&Labels outside of limits

### LINKED ISSUE #249: Scales, breaks and limits and transformations do not work as expected.
```python
g = ggplot(diamonds)
g += geom_point(aes(x='carat', y='price', color='color'), stat='identity')
g += scale_y_continuous(name='price (USD)', 
    trans='log10',
     breaks=[100,1000,10000], labels=['100','1000','10000'])
g += scale_color_discrete(name='color (GIA scale)')
g
```
Expected: A plot.
Actual: Exception, 'Breaks and labels are different lengths. Make sure all the breaks you specify are with in the limits.'

Presumed reason: 100 is outside of the limits.

Ok, let's add some limits:

```g += scale_y_continuous(name='price (USD)', 
    trans='log10',
     breaks=[100,1000,10000], labels=['100','1000','10000'],
                       limits=[100, None])
```
Expected: Plot going from 100 to whatever the auto-training says the limit should be (see changelog for 0.3.0)
Actual: AttributeError: 'int' object has no attribute 'log10'

```g += scale_y_continuous(name='price (USD)', 
    trans='log10',
     breaks=[100,1000,10000], labels=['100','1000','10000'],
                       limits=[100, diamonds.price.max()])
```
At last:
image

The first bug is independent of the transformation, it shows up as long as the break is outside of the limits.

The second one occurs only if the transformation is active.

TyberiusPrime:
This fixes #249.

I've removed the 'store limits as transformed values' from scale_continuous - it wasn't actually true for the default auto-calculated limits, and the unit tests (with default values) actually enforce the limit to be untransformed. 
Arguably, this removes the ability to specify the limits as (big, small), but there was no unit test for this ability.

The second part simply filters the labels the same way the breaks are filtered instead of filtering just the breaks and then complaining that there are the wrong number of labels.


plotnine/scales/scale.py
```diff
@@ -575,9 +574,10 @@ def break_info(self, range=None):
             major = major.compress(np.isfinite(major))
             minor = self.get_minor_breaks(major, range)
 
+        chosen = (range[0] <= major) & (major <= range[1])
         major = major.compress(
-            (range[0] <= major) & (major <= range[1]))
-        labels = self.get_labels(major)
+            chosen)
+        labels = self.get_labels(major, chosen)
```
has2k1: The changes to fix the issue can all be in a single spot.

```python
        else:
            major = major.compress(np.isfinite(major))
            minor = self.get_minor_breaks(major, range

        labels = self.get_labels(major)

        # Filter out breaks outside the limits & corresponding labels
        mask = (range[0] <= major) & (major <= range[1])
        major = [x for x, m in zip(major, mask) if m]
        labels = [x for x, m in zip(labels, mask) if m]
```

plotnine/scales/scale.py
```diff
@@ -662,7 +662,7 @@ def get_minor_breaks(self, major, limits=None):
         else:
             return self.trans.minor_breaks(major, limits)
 
-    def get_labels(self, breaks=None):
+    def get_labels(self, breaks=None, chosen=None):
```
has2k1: Changes to this method are not required.

TyberiusPrime:
Not sure why I'm in confligt, I rebased on master...

TyberiusPrime:
Ok, found my rebase issue.

The falling test cases seem to be related to #266

plotnine/scales/scale.py
```diff
@@ -624,9 +602,10 @@ def break_info(self, range=None):
             major = major.compress(np.isfinite(major))
             minor = self.get_minor_breaks(major, range)
 
-        major = major.compress(
-            (range[0] <= major) & (major <= range[1]))
         labels = self.get_labels(major)
+        mask = (range[0] <= major) & (major <= range[1])
+        major = [x for x, m in zip(major, mask) if m]
+        labels = [x for x, m in zip(labels, mask) if m]
```
has2k1: Okay, I have failed. I tried to think of a way to restrict the changes to this single location, but the other options are worse. There isn't a good way to represent missing values in the python ecosystem, it would have been helpful.

Also, `labels = self.get_labels(major)` has to use the compressed `labels` so that the labelling in `mizani` can make a good decision especially when dealing with large number ranges.

So change this back closer to your original solution but call the extra parameter a `mask` and add a docstring for
`get_labels`.

Thanks for you patience.

TyberiusPrime:
All right, I tried, hope it's closer to what we need. 

The failing test case is because of the removed scale.limits.setter - apparently that new test cases does uncover a case where it was needed? It is not enough to do the transfromation in scale_position_continous.map though - that fails with a different error. 
I need your guidance...

plotnine/scales/scale.py
```diff
@@ -521,25 +518,6 @@ def trans(self, value):
...
-        all computations happen on transformed data. The
-        labeling of the plot axis and the guides are in
-        the original dataspace.
-        """
-        limits = self.trans.transform(value)
-        try:
-            self._limits = np.sort(limits)
-        except TypeError:
-            self._limits = limits
-
```
has2k1: Okay, now that I recall, this code cannot be removed otherwise it will break setting the limits for transformed scales. I think the changes below should be enough.

And maybe this line,

```python
limits = self.trans.transform(value)
```

should deal with None e.g. `[100, None]`.

This should be 2 commits.

1. The case for `[100, None]`.
2. The breaks & labels.

TyberiusPrime:
All right, I finally got around to having another look at this.

Passed all tests locally, let's say what Travis thinks.

has2k1:
@TyberiusPrime, thanks. This issue was kinda stubborn.

PR #251, opened 2019-01-22
Add whitespace to code example.

kingishb:
Hi! Sorry to add a second PR but I went to the docs site to check that the merged PR looked appropriate and it appears that removing whitespace on the first line of the code directive interferes with its rendering. Sorry about that! This should make it visible. You can press view file on the change to verify.


has2k1:
Thanks, I should have caught it!.

PR #254, opened 2019-01-30
introduced scale_position_*(reverse=True)

### LINKED ISSUE #253: scale_y_reverse and transformations does not reverse
Example
 ```python
from plotnine import *
import pandas as pd

df = pd.DataFrame({
    'p': [0.1, 0.2, 0.3, 0.4, 0.5],
    'c': ['red','blue','lightblue','purple','green'],
    'x': ['x'] * 5
    
})
g = ggplot(df)
g += geom_point(aes('x','p', color='c'))
g += scale_color_identity()
g += scale_y_reverse(trans='log10')
g += ggtitle('scale_y_reverse')
g
```

image

Looks identical to scale_y_continuous:
image

TyberiusPrime:
This introduces a parameter 'reverse' to scale_x/y_continuous and scale_x/y_discrete,
which turns the scale around.

This is similar to scale_x/y_reverse, but works on transformed and discrete scales,
and fixes #253.

Both were doable before, just not easily - you'd have to define your own transformation,
and turn your discrete data into a Categorical with reversed categories.

I believe this hits the right abstraction and keeps compatibility with ggplot2.

scale_x/y_reverse are now wrappers around scale_x/y_continuous(reverse=True, *args, **kwargs).

For the continuous case, this is 'easy' 
we simply define a transformation based on the existing one with * -1 at the right places and fiddle with the limits a bit to generate the same breaks.
reverse_transform() function should probably move to mizani though. @has2k1  you're in a better position to make this happen than I am.

For the discrete case this original looked like an easy case of reversing the mapping in scale_position.discrete for both discrete and continuous data, but it turns out that is being called twice from draw and that was relying on the continuous data being identity mapped.
The flow was like this:
 * map discrete -> mapped continuous
 * setup_data of geoms using mapped continuous data
 * map (now continuous) -> no op

This causes issues when you add in a geom that has continuous data - for example a hline,
since that data needs mapping, but only once (or we'd be back at the start again).

I think I've nailed it with a guard in layout.map_position - if not we're lacking a test case to expose it...

I've also included the doc changes suggested in #253 (and of course the doc for reverse=)


has2k1:
Instead of fiddling around with the scales to get the desired result. The solution is reversing the coordinate system. I have fixed the bug in https://github.com/has2k1/plotnine/commit/e6ad834946cc6f8634cc2abc765d3537f504bc82 so that `coord_trans(x='reverse')` now works.

TyberiusPrime:
That's cool. 

Do you want scale_y_reverse to warn if the user sets a trans, to raise, or to create a reversed scale?

I'd also like to amend their annotation to point the user to coord_trans, if such a PR would be welcome

has2k1:
> Do you want scale_y_reverse to warn if the user sets a trans, to raise, or to create a reversed scale?

It needs to be done for a bunch of `scales_*`, I thought about doing it while fixing this issue but could not come up with a nice and tidy way.

PR #255, opened 2019-01-31
rejecting  trans argument on scales with fixed trans

TyberiusPrime:
With regard to our discussion in #254,
this is how I would solve the 'fixed-trans-scales-must-raise-on-trans-paramter != their-trans' goal.

All it needs is a mixin base class - the scales were using multiple inheritance already (scale_x_datetime),
so no additional magic needs to be understood (unlike e.g. a solution with a class decorator) and the boilerplate is minimal.

Not sure if you want the base class with @document.

### DIFF
--- a/plotnine/scales/scale.py
+++ b/plotnine/scales/scale.py
@@ -718,8 +718,26 @@ class scale_continuous(scale):
         return labels
 
 
+class scale_continuous_fixed_transformation(scale_continuous):
+    """Base class/mixin for scales that have one fixed transformation,
+    which may not be overwritten by the user
+    e.g. scale_x_log10
+    """
+
+    def __init__(self, *args, **kwargs):
+        if 'trans' in kwargs and kwargs['trans'] != self._trans:
+            name = self.__class__.__name__
+            xy = 'x' if '_x_' in name else 'y'
+            raise ValueError(("%s is fixed to transformation %s "
+                              "- use scale_%s_continuous instead") %
+                             (name,
+                              self._trans,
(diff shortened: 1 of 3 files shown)


has2k1:
I was reluctant to introduce a mixin class because, I wanted to show a warning and a mixin would be too much just for that function. I think only the documentation needs to change.

TyberiusPrime:
I believe warnings are for 'this might be wrong or not, the lib can't tell'-situations, while
obviously wrong (as in "you have told me contradictory things to do") should raise an Exception.

I can see two other straight forward implementations, either with a class decorator (magic), or a 
flag on the transformed-scale classes (e.g. use '_fixed_trans' instead of _trans) which scale_continuous then interprets as 'warn/raise on trans'.




(Philosophically, having two ways to define transformations, scale_continous(trans=) and scale_trans is unpythonic - but nothing we can do about that).

has2k1:
> use '_fixed_trans' instead of _trans

That would work, but I think there is a better way.

TyberiusPrime:
A separate flag?


You might get away with having the test in scale_continuous if _trans != Identity (or waiver...) But it is not clear and poses a restriction for all child classes  not just those that opt in.

has2k1:
Thanks for raising and exploring this issue, I have resolved it in https://github.com/has2k1/plotnine/commit/0c9f85b1eb183dd229e5c770d8416aedab4926a9 with a warning when the `trans` for non-identity scale is set to a different transformation.

PR #257, opened 2019-01-31
stats check for unknown arguments when used in a geom

TyberiusPrime:
Bug
```python
g += geom_point(aes('group','y', color='group'),
                stat = stat_summary(funy=np.mean),                
               )```
Did not raise an exception (note misspelling of fun_y).

```
g += geom_point(aes('group','y', color='group'),
                stat = stat_summary(),                
                funy=np.mean
               )
```
does though.

This extends the geoms error checking to also check the stat's arguments.
We can't do it in the stat.__init__ since it is ok to pass geom arguments here,
for example if using the stat as a standalone layer, so we have to do it in the geom's verify_arguments.

### DIFF
--- a/plotnine/geoms/geom.py
+++ b/plotnine/geoms/geom.py
@@ -359,7 +359,8 @@ class geom(metaclass=Registry):
         """
         Verify arguments passed to the geom
         """
-        unknown = (kwargs.keys() -
+        geom_stat_args = kwargs.keys() | self._stat._kwargs.keys()
+        unknown = (geom_stat_args -
                    self.aesthetics() -                # geom aesthetics
                    self.DEFAULT_PARAMS.keys() -        # geom parameters
                    self._stat.aesthetics() -          # stat aesthetics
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plotnine/geoms/geom.py [hunk 1]
```diff
@@ -359,7 +359,9 @@ def _verify_arguments(self, kwargs):
         """
         Verify arguments passed to the geom
         """
-        unknown = (kwargs.keys() -
+        seen = set(kwargs.keys())
+        seen.update(self._stat._kwargs.keys())
+        unknown = (seen -
```
has2k1: Can you change it to

```python
        geom_stat_kwargs = kwargs.keys() ^ self._stat._kwargs.keys()
        unknown = (geom_stat_kwargs -
                   ...
```

TyberiusPrime: Not sure I follow - why would you want the symmetric difference of those two instead of their union? ( ^ vs | )

has2k1: Sorry, it should be It should be union `|`.

TyberiusPrime:
Is this ok now?

has2k1:
Yes. Thanks, I have merged it in https://github.com/has2k1/plotnine/commit/33626528193e408feeaae7be44564e663344f0ee.

Issue #276, opened 2019-05-05
new deprecations warning with Matplolib-3.1.rc


stonebig:
````
C:\WinP\bd37\buPs2\WPy64-3730b2\python-3.7.3.amd64\lib\site-packages\plotnine\scales\scale.py:93: MatplotlibDeprecationWarning: 
The iterable function was deprecated in Matplotlib 3.1 and will be removed in 3.3. Use np.iterable instead.
  if cbook.iterable(self.breaks) and cbook.iterable(self.labels):
C:\WinP\bd37\buPs2\WPy64-3730b2\python-3.7.3.amd64\lib\site-packages\plotnine\utils.py:553: MatplotlibDeprecationWarning: 
The iterable function was deprecated in Matplotlib 3.1 and will be removed in 3.3. Use np.iterable instead.
  return cbook.iterable(var) and not is_string(var)
````

jorisvandenbossche:
Are there plans for a new release? (the warnings are quite annoying ..)

PR #277, opened 2019-05-05
matplolib-3.1.0 deprecation warning

stonebig:
solves https://github.com/has2k1/plotnine/issues/276

PR #283, opened 2019-06-16
Matplotlib-3.1.0 deprecation fix

stonebig:

PR #285, opened 2019-06-25
Further fix issue 276.

blaiseli:
This is to try to fix https://github.com/has2k1/plotnine/issues/276
There were still at least one case of warning-triggering code left after https://github.com/has2k1/plotnine/pull/277


has2k1:
This issue has been fixed. [See](https://github.com/has2k1/plotnine/pull/283).

PR #301, opened 2019-08-19
fix: annotation_logticks & any discrete scale

TyberiusPrime:
Adding annotation_logticks to a plot with a discrete scale fails with an exception:
```python
/project/code/plotnine/plotnine/geoms/annotation_logticks.py in _check_log_scale(base, sides, panel_params, coord)
     55         x_scale = panel_params.x.scale
     56         y_scale = panel_params.y.scale
---> 57         x_is_log = is_log(x_scale.trans)
     58         y_is_log = is_log(y_scale.trans)
     59         if isinstance(coord, coord_flip):

AttributeError: 'scale_x_discrete' object has no attribute 'trans'
```

This is independent of the sides argument.

This PR simply adds a check for the scale having a .trans attribute in is_log, and later on, if the base has not been set by the user (we didn't test for this combination of non-log axis and no base specified), 
default to base 10 (arbitrary, but we're in 'warning, your scale in not log'-territory anyway).

### DIFF
--- a/plotnine/geoms/annotation_logticks.py
+++ b/plotnine/geoms/annotation_logticks.py
@@ -47,21 +47,27 @@ class _geom_logticks(geom_rug):
         out : tuple
             The bases (base_x, base_y) to use when generating the ticks.
         """
-        def is_log(trans):
+        def is_log(scale):
+            if not hasattr(scale, 'trans'):
+                return False
+            trans = scale.trans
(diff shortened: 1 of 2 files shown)


has2k1:
@TyberiusPrime, nice. Thank you.

I want to have v0.6.0 released this week.

PR #302, opened 2019-08-20
fix: IndexError if values exceeded scale_discrete palette size

TyberiusPrime:
This plot would fail with an IndexError in map (and not just a warning for omitted values due to palette limits):

```python
  df = pd.DataFrame({
        # not that it's the second c that triggered a bug in scale_discrete.map
        'x': pd.Categorical(['a', 'b', 'c', 'c']),
        'y': [0, 1, 2, 2]})
    p = (ggplot(df, aes('x', 'y', color='x', shape='x'))
         + geom_jitter()
         + scale_color_manual(['red', 'blue'])
         ).save('test.png')
```

This PR fixes it by replacing a complicated numpy 'insert NaNs at indices' solution with a call to pandas.Series.reindex() which helpfully replaces non-existent indices with the NaNs we need.

### DIFF
--- a/plotnine/tests/test_scale_internals.py
+++ b/plotnine/tests/test_scale_internals.py
@@ -7,7 +7,7 @@ import pytest
 
 from plotnine import ggplot, aes, geom_point, expand_limits, theme
 from plotnine import geom_col, lims, element_text, annotate
-from plotnine.scales import scale_color
+from plotnine.scales import scale_color, scale_color_manual
 from plotnine.scales import scale_identity
 from plotnine.scales import scale_manual
 from plotnine.scales import scale_xy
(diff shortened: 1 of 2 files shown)


plotnine/tests/test_scale_internals.py
```diff
@@ -645,3 +645,16 @@ def test_layer_with_only_infs():
          )
     p = p.build_test()
     assert isinstance(p.scales.get_scales('x'), scale_x_discrete)
+
+
+def test_discrete_scale_exceeding_maximum_number_of_values():
+    df = pd.DataFrame({
+        # not that it's the second c that triggered a bug in scale_discrete.map
+        'x': pd.Categorical(['a', 'b', 'c', 'c']),
```
has2k1: Use `['c', 'a', 'c', 'b', 'c']`. If the logic, ever changes this is trickier.

has2k1:
Also, keep the commit title [50 characters or less](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html)

TyberiusPrime:
I fixed the commit title and switched to your test suggestion.

As for the actual code, I did some profiling:
image

pandas is the one I proposed initially, list_comp is your suggestion, 
clip is the one I've pushed now.

I'm currently calculating with sample size = 100k...

TyberiusPrime:
Ok, 100k results are in - clip still beats pandas at that size, so let's stick with that.


image

has2k1:
Thanks. Revealing profiling results, kind of what I would have expected.

PR #304, opened 2019-08-22
feat: blended edge stripes in annotation_stripes

TyberiusPrime:
This adds the fill_range option = 'blend' to annotation_stripes which turns this
image

into this
image

which at least in my use case is what I actually want for discrete scales.


plotnine/geoms/annotation_stripes.py
```diff
@@ -20,10 +20,11 @@ class annotation_stripes(annotate):
     fill : list-like
         List of colors for the strips. The default  is
         `("#AAAAAA", "#CCCCCC")`
-    fill_range: bool
-        If True, the more stripes are created to cover the
+    fill_range: bool or 'blend'
```
has2k1: I think it is better to have two parameters

```
fill_range: bool
fill_range_how: 'auto' | 'cycle' | 'nocycle'
```

When `fill_range` is True, then `fill_range_how` decides. It makes sense to me for the default filling for continuous scales to keep cycling the colors, and for the discrete scales to maintain the same color. I do not know if there are better names than 'cycle' and 'nocycle' (what you referred to as 'blend').

TyberiusPrime: I don't agree with splitting it into two arguments - it's one enum that defines how to handle the edges of the range. We might not have False|True|"blend" (mixture of types) 
but three strings instead:

* no -> don't fill the range
* stripes -> put the next stripe color in place (=cycle)
* no stripes -> visually extend the first/last stripe (=no cycle, blend)

I personally don't have much use for stripes, but it's in the library now (as True)

TyberiusPrime: (I wish python had first class enums)

TyberiusPrime:
Love it, thanks!

PR #315, opened 2019-09-11
Fix: annotation_stripes with single stripe

TyberiusPrime:
annotation_stripes crashes when there is only a single break:

```python
       n_stripes = len(breaks)
        diff = np.diff(breaks)
>       step = diff[0]
E       IndexError: index 0 is out of bounds for axis 0 with size 0
geoms/annotation_stripes:82: IndexError
```

This let's it draw a single stripe instead.

There is a slight discrepancy in that multiple stripes
leave space at the sides, and the single one does not
(as if fill_range was set to True),
but I believe this is actually upstream of annotation_stripes,
when axis extends are calculated.


plotnine/tests/test_annotation_stripes.py
```diff
@@ -76,3 +76,13 @@ def test_annotation_stripes_fill_direction_extend():
...
 
     assert p == 'annotation_stripes_fill_direction_extend'
+
+
+def test_annotation_stripes_single_stripe():
+    p = (ggplot(df.assign(x=10))
+         + annotation_stripes()
+         + geom_point(aes('factor(x)', 'y'))
+         + geom_vline(xintercept=[0.5, 1.5, 2.5, 3.5])
+         )
```
has2k1: Can you use different colours for the stripe. For a single stripe, the default colour (grey) looks so much like the plot background. Does `geom_vline` have any purpose here?, in other tests the lines mark were we expect the boundaries of the stripes to be.

PR #316, opened 2019-09-11
fix: geom_violin and very small n

TyberiusPrime:
geom_violin crashed if you only had x values with
one or two (unique) data points. If other x
values with more data points where present,
it would not show anything for the ones
with only one or two data points. The later
behavior was apparently inherited from ggplot2.

This fixes both by showing singular values
as a line (just like a boxplot does) and two values as a box before
going into the 'proper' density estimate
starting at three (unique) data points.

While this is arguably not ideal,
it is visually distinct from a full density estimate,
and *does not crash*.

I hate my QC plots crashing just because the 
data is a bit degenerate.

Example:
```python
from plotnine import *
df = pd.DataFrame({
        'x': ['A'] + ['B', 'B'] + ['C'] * 20 +
             ['D', 'D', 'D', 'E', 'E'] + ['F'] * 20 + ['G', 'G', 'G'],
        'y': [0, 100, 100] + [200] * 20 +
             [0, 50, 50] + [100, 150] + [150, 200] * 10 + [10, 20, 60],
})
p = (ggplot(df, aes('x')) +
     geom_violin(aes(y='y')))
p
```
used to generate
image

and now generates 

image

(note that D and E have only two distinct values. We should not treat 
[10,20] different from [10,10, 20,20] in this plot, but the underlying stat_density totally does (it also only checks for the number of data points, not the number of unique values). So I fixed it to always go for the degenerate case.


has2k1:
> We should not treat [10,20] different from [10,10, 20,20]

I do not think that is right. The density estimation algorithm sums up to `N` (the number of points), *every point* contributes a "kernel" and the summation of those kernels is the KDE.

Given KDE theory (with a gaussian kernel), if you think about [10, 20] vs [10, 10, 20, 20], the relative height of the peaks has to be the same but the valley between the two peaks has to be deeper for [10, 10, 20, 20]. What it means is that you have collected more data and 15 has still not showed up!, so if you are using the KDE to generate data you should not be predicting 15 with the same probability for the two cases. The KDE approximation of the underlying distribution gets more accurate as the data increases.

If you test for uniqueness and generate some PDF based on that, you are essentially creating your own kernel!

I have fixed the crashing issue, but what I see as missing is that the bandwidth parameter (which controls the width of the kernel function) is not part of the API.

PR #325, opened 2019-10-21
Turn ha and va into aesthetics for geom_text()

jeroenjanssens:
While recreating the following figure of [Section 28.3 from R for Data Science](https://r4ds.had.co.nz/graphics-for-communication.html#annotations):

I noticed that `ha` and `va` are not aesthetics but parameters. According to the documentation, this is because of a limitation in Matplotlib (see https://github.com/matplotlib/matplotlib/pull/1181). Unless I fail to fully understand the issue, I believe that this small change fixes it. At least I was able to reproduce the above figure with the following code:

```python
from plotnine import *
import pandas as pd
from itertools import product

has = ["left", "center", "right"]
vas = ["top", "center", "bottom"]
xs = [0, 0.5, 1]
ys = [1, 0.5, 0]

df = pd.DataFrame([{"x": xs[x],
                    "y": ys[y],
                    "ha": has[x],
                    "va": vas[y],
                    "label": f"ha=\"{has[x]}\"\nva=\"{vas[y]}\""}
                   for x, y in product(range(3), repeat=2)])

ggplot(df, aes("x", "y")) +\
geom_point(colour="grey", size=5) +\
geom_point(size=0.5, colour="red") +\
geom_text(aes(label="label", ha="ha", va="va"), size=14) +\
labs(x=None, y=None)
```

download (4)

Please forgive me for not yet taking care of the deprecated terms "hjust" and "vjust". I wasn't able to quickly figure out how to treat those as aliases in the case of aesthetics.

### DIFF
--- a/plotnine/geoms/geom_text.py
+++ b/plotnine/geoms/geom_text.py
@@ -38,10 +38,6 @@ class geom_text(geom):
         Font weight.
     fontstyle : str (default: normal)
         Font style. One of *normal*, *italic* or *oblique*
-    ha : str (default: center)
-        Horizontal alignment. One of *left*, *center* or *right.*
-    va : str (default: center)
-        Vertical alignment. One of *top*, *center* or *bottom.*
     nudge_x : float (default: 0)
         Horizontal adjustment to apply to the text
     nudge_y : float (default: 0)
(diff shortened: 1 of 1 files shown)


has2k1:
This PR also raises the issue that the aesthetics are not documented (with descriptions). So far, I think all the aesthetics have been self-descriptive, but `va` and `ha` are not.

has2k1:
> This PR also raises the issue that the aesthetics are not documented (with descriptions). So far, I think all the aesthetics have been self-descriptive, but va and ha are not.

I was wrong. There is a provision for documenting them as some `stats` do. I had forgotten about it!

has2k1:
@jeroenjanssens. Thanks. I will handle the documentation and the old parameters names.

PR #361, opened 2020-03-03
Fix #359. `geom_map` uses incorrect color/fill indexing, is incompatible with faceting

### LINKED ISSUE #359: `geom_map` uses incorrect color/fill indexing, is incompatible with faceting
### Overview

Hey there!

When attempting to add `facet_wrap` to a plot using `geom_map`, I ran into this exception:

   IndexError: positional indexers are out-of-bounds

In the Traceback, it is this section of `geom_map.py` that's the culprit:

https://github.com/has2k1/plotnine/blob/246252c6b2e84a8e7dc83ee8661972496d22a1ab/plotnine/geoms/geom_map.py#L98-L105

After a bit of messing around, I discovered two bugs in the code.

### Bug 1

`cidx` and `fidx` are indices produced through subsetting and therefore have in-index values, but the lines where `color` and `fill` columns are set use `iloc`, which uses a positional indices. This works for non-faceted plots (and the first facet) because the index is created as an Int64Index, where positional index is the same as the associated index value, but not for other facets, where they do not match.

One potential fix is to use `.loc` instead of `.iloc` and ignore the `_loc` function:

        cidx = data.index[data['color'].isnull()]
        fidx = data.index[data['fill'].isnull()]
        data.loc[cidx, 'color'] = 'none'
        data.loc[fidx, 'fill'] = 'none'

### Bug 2

I thought this would fix everything, but the same type of issue occurs just after when figuring out the geometry type.

Specifically, `data.geometry.loc[0].geom_type` looks up the index value of 0, rather than the positional value, and facets other than the first one won't have this value. The easiest fix is to use `iloc`: `data.geometry.iloc[0].geom_type`.

### Summary

Replacing the current lines 98-105 with this code should fix the issue. Let me know if it's better for me to create a pull request:

        _loc = data.columns.get_loc
        cidx = data.index[data['color'].isnull()]
        fidx = data.index[data['fill'].isnull()]
        data.loc[cidx, 'color'] = 'none'
        data.loc[fidx, 'fill'] = 'none'
        data['fill'] = to_rgba(data['fill'], data['alpha'])

        geom_type = data.geometry.iloc[0].geom_type

`_loc` is also no longer used in this code and could be removed.

nbolten:
This pull request implements the changes suggested in #359, including the removal of the _loc variable. It also removes the related `cidx` and `fidx` variables, since they're only used once and their expressions can fit in concise, readable one-liners.

### DIFF
--- a/plotnine/geoms/geom_map.py
+++ b/plotnine/geoms/geom_map.py
@@ -95,14 +95,11 @@ class geom_map(geom):
         if not len(data):
             return data
 
-        _loc = data.columns.get_loc
-        cidx = data.index[data['color'].isnull()]
-        fidx = data.index[data['fill'].isnull()]
-        data.iloc[cidx, _loc('color')] = 'none'
-        data.iloc[fidx, _loc('fill')] = 'none'
+        data.loc[data['color'].isnull(), 'color'] = 'none'
+        data.loc[data['fill'].isnull(), 'fill'] = 'none'
         data['fill'] = to_rgba(data['fill'], data['alpha'])
 
-        geom_type = data.geometry[0].geom_type
+        geom_type = data.geometry.iloc[0].geom_type
         if geom_type in ('Polygon', 'MultiPolygon'):
             data['size'] *= SIZE_FACTOR
             patches = [PolygonPatch(g) for g in data['geometry']]


has2k1:
Thanks @nbolten. This has been [merged](https://github.com/has2k1/plotnine/commit/67ccdf4bb86af1a479bf47441c4af7f183609e08). I have also added tests.

PR #385, opened 2020-04-17
fix: Alpha respected by polygon lines

TyberiusPrime:
Many geoms only partially support alpha in the sense that their borders are always at alpha=1.

This fixes this for anything using geom_polygon plus geom_boxplot.

Example:
image

This is a 'minimal' change that makes alpha work on both the fill and the lines.
Doing it this way gives the user the option to draw the lines at alpha = 1 by repeating the geom with fill=None, increasing the number of possible plots.

The other option would have been to introduce 'color_alpha' at every geom where you'd consider it (since the color comes from... color ;)).

ggplot2 bug tracker suggests they fixed alpha to affect both at one point: https://github.com/tidyverse/ggplot2/issues/1371 but this seems to have regressed?

Anyhow, this patch enables me to draw graphs like this (note the subtle glow):
image


has2k1:
I kind of recall the reason for this being mentioned in one of the ggplot2 google groups. This was a long while back.

TyberiusPrime:
I agree that the clean way would be to have two mappings.
But either way it means we must deviate from the ggplot API a bit. Now I believe additions are ok...

TyberiusPrime:
I'm willing to change the implementation to two mappings. Just tell me what to call it.

PR #409, opened 2020-06-17
fix aes argument typehint according to PEP0484

FFroehlich:
https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values

### DIFF
--- a/plotnine/aes.py
+++ b/plotnine/aes.py
@@ -38,7 +38,7 @@ class aes(dict):
         x aesthetic mapping
     y : expression | array_like | scalar
         y aesthetic mapping
-    **kwargs : dict
+    **kwargs : str
         Other aesthetic mappings
 
     Notes


has2k1:
I see the mix-up. `kwargs: dict` in the docstring means that kwargs is a `dict`, it does not say anything about the type of keys or values in the `dict`. The libraries in the scientific python ecosystem document keyword parameters the same way. This is different from pep-0484 were you give the type hint of the keyword parameters.

I wonder, is an IDE is interpreting `kwargs: dict` as a type hint for the values?

FFroehlich:
> I see the mix-up. `kwargs: dict` in the docstring means that kwargs is a `dict`, it does not say anything about the type of keys or values in the `dict`. The libraries in the scientific python ecosystem document keyword parameters the same way. This is different from pep-0484 were you give the type hint of the keyword parameters.
> 
> I wonder, is an IDE is interpreting `kwargs: dict` as a type hint for the values?

Correct, PyCharm is interpreting this as typehint for the values. Do you believe this is an issue on their end? I see that pep-0484 doesn't really apply here, yet I would argue that the same type of type documentation is plausible here.

has2k1:
Then, do you know if `dict[str, object]` would be treated better?

FFroehlich:
> Then, do you know if `dict[str, object]` would be treated better?

It is treated the same, so it interprets that the values should be dicts.

PR #417, opened 2020-07-20
Change hasattr to callable

fcatus:
It is unreliable to use `hasattr(object, '__call__')` to check if an object is callable as it can give wrong results if object implements custom `__getattr__` or its `__call__` is itself not callable. It is recommended to use the built-in `callable` instead.


has2k1:
Thanks for catching those lines. You accidentally committed .DS_Store files!

fcatus:
Yikes, sorry!  I fixed it and sent a new PR.  I also added .DS_Store to `.gitignore` for the convenience of future MacOS collaborators :)!

Issue #436, opened 2020-09-07
Set the font family in `rcParams` as default for some themes


Gedevan-Aleksizde:
Hello, thank you for this great package!

Some themes' default font family is set as DejaVu Sans, which has large coverage for unicode -- Extended Latin, Cyrricil, Arabic, and so on. But this font doesn't cover CJK (Chinese, Japanese and Korean) fonts.
I usually set a CJK font in `rcParams` but these themes' default values overwrite it.  I want to avoid to set the same font again. Why don't you change the default value to `None`

has2k1:
In the next version of plotnine you will be able to set a default font for all your themes with

```python
from plotnine.options import set_option
set_option('base_family',  'My Font')
```

PR #437, opened 2020-09-07
Set the font family in `rcParams` as default for some themes

Gedevan-Aleksizde:
Please refer to #436

Some themes' default font family is set as DejaVu Sans, which has large coverage for unicode -- Extended Latin, Cyrricil, Arabic, and so on. But this font doesn't cover CJK (Chinese, Japanese and Korean) fonts. So I propose changing the default value to `None`

### DIFF
--- a/plotnine/themes/theme_538.py
+++ b/plotnine/themes/theme_538.py
@@ -16,7 +16,7 @@ class theme_538(theme_gray):
     base_family : str, optional
         Base font family.
     """
-    def __init__(self, base_size=11, base_family='DejaVu Sans'):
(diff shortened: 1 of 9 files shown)


has2k1:
Rather than using rcParams, a better (unlikely to break in the future) solution is to create an option for the `base_family` [within plotnine](https://github.com/has2k1/plotnine/blob/master/plotnine/options.py), so that themes that take `base_family` as a parameter use it.

```python
from plotnine.options import get_option

if base_family is None:
    base_family = get_option('base_family')
```
Then you would use `set_option('base_family', 'cjk_font'`).

PR #440, opened 2020-09-16
Handle change in domain check behaviour in statsmodels 0.12.

jsspencer:
statsmodels 0.12 fixed a bug where evaluating points outside of the
domain in density estimation now returns np.nan instead of raising a
ValueError. Detect both cases and fallback to estimating the density
point-wise.

Context: https://github.com/statsmodels/statsmodels/pull/6547


has2k1:
@jsspencer, thank you for the fix.

Issue #460, opened 2020-11-19
Allow grouped DataFrames as input to ggplot


machow:
Hey! Currently ggplot can take arguments in a number of formats, but grouped DataFrames aren't accepted:

```python
import pandas as pd
from ggplot import ggplot, aes

df = pd.DataFrame({'g': [1,1,2], 'x': [3,4,5]})

# works
ggplot(df, aes("x"))

# errors
ggplot(df.groupby('g'), aes("x"))

# also errors
df.groupby('g') >> ggplot(aes("x"))
```

It looks like supporting pandas.core.groupby.DataFrameGroupBy might only need a few small changes! I've opened a PR with a basic implementation and tests, but am likely missing edge cases, like passing data to geoms?

WDYT? The one potential downside is that AFAIK detecting a DataFrameGroupBy requires reaching into `pandas.core` (though, could detect an attribute instead?).

PR #461, opened 2020-11-19
ENH: ggplot calls accept grouped DataFrames

machow:
addresses #460, but I'm guessing there may be other cases to handle (like passing data to a geom?)

### DIFF
--- a/plotnine/tests/test_ggplot_internals.py
+++ b/plotnine/tests/test_ggplot_internals.py
@@ -82,7 +82,7 @@ def test_ggplot_parameters_grouped():
 
     assert p.data is df
     assert p.mapping == aes('x')
-    
+
     p = ggplot(data=df, mapping=aes('x'))
     assert p.data is df
     assert p.mapping == aes('x')
--- a/plotnine/tests/test_ggplot_internals.py
+++ b/plotnine/tests/test_ggplot_internals.py
@@ -77,6 +77,17 @@ def test_ggplot_parameters():
         ggplot([1, 2, 3], aes('x'))
 
 
+def test_ggplot_parameters_grouped():
+    p = ggplot(df.groupby('x'), aes('x'))
+
+    assert p.data is df
+    assert p.mapping == aes('x')
+    
+    p = ggplot(data=df, mapping=aes('x'))
+    assert p.data is df
+    assert p.mapping == aes('x')
+
+
 def test_data_transforms():
     p = ggplot(aes(x='x', y='np.log(y)'), df)
     p = p + geom_point()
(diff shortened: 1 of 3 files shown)


plotnine/tests/test_ggplot_internals.py
```diff
@@ -77,6 +77,17 @@ def test_ggplot_parameters():
...
 
+def test_ggplot_parameters_grouped():
+    p = ggplot(df.groupby('x'), aes('x'))
+
+    assert p.data is df
+    assert p.mapping == aes('x')
+    
+    p = ggplot(data=df, mapping=aes('x'))
+    assert p.data is df
+    assert p.mapping == aes('x')
```
has2k1, line 88: I am wondering why not `p.mapping == aes('x', group='x')`. Would it be overly presumptuous to turn the pandas groups into a mapping to the group variable?

plotnine/utils.py
```diff
@@ -962,6 +963,11 @@ def data_mapping_as_kwargs(args, kwargs):
         raise PlotnineError(msg.format(duplicates))
     return kwargs
 
+def ungroup(data):
+    if isinstance(data, DataFrameGroupBy):
+        return data.obj
+
+    return data
```
has2k1, line 970: Missing docstring.

plotnine/utils.py
```diff
@@ -1001,7 +1007,7 @@ def order_as_mapping_data(*args):
             "but I got {}.".format(n)
         )
 
-    mapping, data = args
+    mapping, data = map(ungroup, args)
```
has2k1, line 1010: Something else other than `map`, a generator comprehension maybe!

has2k1 (commented):
I see how not having this can irk a `suiba` user.

plotnine/tests/test_ggplot_internals.py
machow: Ah, good call--it looks like this is what ggplot does...

https://github.com/tidyverse/ggplot2/blob/d700a0b99ad9776bd9fcc17ff568c182c458c3ee/R/fortify.r#L37

plotnine/utils.py
machow: Ah, thanks--added

plotnine/tests/test_ggplot_internals.py
has2k1: I don't know what ggplot does, but I had this functionality
https://github.com/has2k1/plotnine/blob/2b0ccdba4ed385eacaedfbb1d410ec55d10bbbfa/plotnine/ggplot.py#L60-L63
for plydata. 

However, there is an inconsistency which I realised because of this PR. It only works when passed through the `ggplot()` call and I do not think it is worth it to add more code. So it may be dropped entirely.

has2k1:
Thanks.

PR #462, opened 2020-11-20
Add 'ngroup' to stat_bin

CarlosGrohmann:
add `ngroup` stat (number of points per group) to `stat_bin`


has2k1:
@CarlosGrohmann great.  thank you for seeing it through.

Issue #428, opened 2020-08-08
Doesn't work if the plot is not the last thing in the cell


impact27:
If I run:
```python
from plotnine import ggplot, geom_point, aes, stat_smooth, facet_wrap
from plotnine.data import mtcars

(ggplot(mtcars, aes('wt', 'mpg', color='factor(gear)'))
+ geom_point()
+ stat_smooth(method='lm')
+ facet_wrap('~gear'))

a = 1 + 1
```
in a jupyter notebook, the plot won't be shown. It only works if the plot is at the end of the expression. 
This causes issue https://github.com/spyder-ide/spyder/issues/13464

has2k1:
Jupyter notebook automatically prints the last object cell and in plotnine the plot drawn by printing the ggplot object. So to get automatic plotting the ggplot object has to be the last object in the cell. Otherwise you have to explicitly print it.

```python
from plotnine import ggplot, geom_point, aes, stat_smooth, facet_wrap
from plotnine.data import mtcars

p = (ggplot(mtcars, aes('wt', 'mpg', color='factor(gear)'))
 + geom_point()
 + stat_smooth(method='lm')
 + facet_wrap('~gear')
)

print(p) 

a = 1 + 1
```

Jupyter treats matplotlib with preference way by printing any figures created in the cell. If you want plotnine to leverage that preference, then you need to explicitly draw the plot which creates a matplotlib figure in the background is noticed by jupyter.

```python
(ggplot(mtcars, aes('wt', 'mpg', color='factor(gear)'))
 + geom_point()
 + stat_smooth(method='lm')
 + facet_wrap('~gear')
).draw()

a = 1 + 1
```

It is better to print the object.

rhalbersma:
I would be nice if ggplot objects had a `__str__` method that draws the plot without the `<ggplot:hash>` annotation. For pretty printing, users could then write `print(p)` and for debugging `print(repr(p))` would still be available.

has2k1:
@rhalbersma, please submit a separate issue.

PR #465, opened 2020-12-08
Fix processing of weight aesthetic for boxplots

joshhartmann11:
Issue #428 revealed `geom_boxplot` ignores the weight aesthetic.  ~~The solution assumes that the weight column corresponds to the integer number of samples.~~ The solution uses a weighted percentile method to allow for non-integer weights

MRE:
```
import pandas as pd
import plotnine as p9

df = pd.DataFrame(dict(y=[1,10], weight=[1,10]))

plot = p9.ggplot(df, p9.aes(x=1, y='y', weight='weight')) + p9.geom_boxplot()
```
MRE Before:
test_before

MRE After:
test_after

Fixes issue #428


has2k1:
Thanks for trying this out. I considered solving it that way too but I concluded that it is a very restricted solution if it does not permit floating point weights; yet what we really need is a way to compute weighted quantiles.

I was going to tackle this issue it week before the upcoming release (latest next week).

This is what I had noted down.

---

- Work off of https://stackoverflow.com/a/29677616/832573 to calculate weighted quantiles `(q1, med, q3)`
  https://www.statsmodels.org/stable/generated/statsmodels.stats.weightstats.DescrStatsW.html
  does compute weighted quantiles but it returns the exact values of input. It does not
  interpolate.
- Compute the rest of the values from the quantiles

```python
q1 =
q3 =
med =
iqr = q3 - q1
whislo = q1 - 1.5*(q3 - q1)
whishi = q3 + 1.5*(q3 - q1)
outliers = data[data.x < 'whislo' or data.x > 'whishi']
```

joshhartmann11:
I've updated the code to support non-integer weights as suggested using a weighted percentile generalisation found on [wikipedia](https://en.wikipedia.org/wiki/Percentile).  The percentiles are then calculated and linearly interpolated in the same way as matplotlib's `boxplot_stats`

plotnine/tests/test_geom_boxplot.py
```diff
@@ -21,7 +26,8 @@ class TestAesthetics:
          geom_boxplot(aes(y='y'), size=2) +
          geom_boxplot(df[:2*m], aes(y='y+25', fill='x'), size=2) +
          geom_boxplot(df[2*m:], aes(y='y+30', color='x'), size=2) +
-         geom_boxplot(df[2*m:], aes(y='y+55', linetype='x'), size=2)
+         geom_boxplot(df[2*m:], aes(y='y+55', linetype='x'), size=2) +
+         geom_boxplot(df[:2*m], aes(y='y+55', weight='weight'), size=2)
```
has2k1, line 30: Can you create a separate test case specifically for the weight one that is intuitive to the reader (at least with some concentration). For example using a dataframe like this:

```python
# Data with two groups,
# - 'a' has no weights
# - 'b' has weights that should make the resulting
#    boxplot same as that of 'a'
df = pd.DataFrame({
    'x': list('a'*10 + 'b'*5),
    'y': np.hstack([
        [1, 2, 2, 3, 3, 3, 4, 4, 4, 4, 20],
        [1, 2, 3, 4, 20]  # 20 should be an outlier
    ]),
    'w': np.hstack([
        np.ones(10),
        [1, 2, 3, 4, 1]
    ])
})
```
I actually don't know if it works, but I hope you get what I am trying to do!

joshhartmann11:
Thank-you for your continued patience.

The weighted percentile method complicated things a bit, for example:

For the set of numbers with constant weight: `[1 ,2, 2, 3, 3, 3, 4, 4, 4, 4, 15]`, the median is clearly `3`

For the set of numbers: `[1,2,3,4,15]` with weights: `w_n = [1,2,3,4,1]`, the weighted median is calculated in the following way:

`S_n = cumulative sum of weights = [1, 3, 6, 10, 11]`
`S_N = sum of weights = 11`
`C = variation of method, 1 is used in numpy = 1`

Per wikipedia:
`p_n = The percentile of the n'th position = (S_n - C * w_n) / (S_N + (1 - 2 * C) * w_n) = [0, 0.11111, 0.375, 0.85714, 1]`

Interpolating the 50th percentile:
`perc_50 = 3.25925926`

As you can see, the methods do not yield the same results. From my research online, there doesn't appear to be a solution for floating point weighted percentiles that also stays consistent with the idea that a weight of `n` means there are 'n' of that sample.

I hope this makes sense, I'm not sure what decision should be made around this.  ggplot2 uses quantile regression, I'll look into that next.

has2k1:
Okay, the equivalent boxplots that I expected with the test case are not possible. That makes sense as per the Wikipedia comment *"There is no standard function for a weighted percentile"*. I think this is because there are many options for the prediction scheme from weight space to data space. Even with linear interpolation you can definitely choose any C in `[0, 1]`. I guess you could even pull out linear regression and fancier schemes to exploit a relationship between the data and the weights!

In light of the above realisation, that median `perc_50 = 3.25925926` makes more sense than if it were `3`. But, what really matters is that x% of the weight lies below than the calculated value. 

You can evaluate quantile regression if you feel like it; otherwise, you can put a comment in the test case why the boxplots are different.

With that, I have learnt that the unweighted median is a discrete problem (yes you interpolate/average when the number of samples is odd), while weighted median is a continuous problem (though you could also round to the nearest value in the data).

has2k1:
@joshhartmann11, Thanks very much taking this one and seeing it through.

Issue #439, opened 2020-09-10
geom_pointdensity


zacmon:
Does a feature [like this]( https://github.com/LKremer/ggpointdensity) made for ggplot2 exist in plotnine where a scatter plot has density information? Presently, I couldn't find any existing documentation which mirrored this plot behavior.

has2k1:
Plotnine does not have this geom. It is well motivated in the description, I would consider PR if one is submitted.

PR #468, opened 2020-12-18
Add geom_pointdensity

joshhartmann11:
Here is my go at making geom_pointdensity after #439.  In stat_pointdensity I used the KDE tools already present to calculate the density at each point rather than using "n neighbours"  It seems like [LKremer at the ggplot point density implementation](https://github.com/LKremer/ggpointdensity) is using this method as well (Although their readme talks a lot about "n neighbours").

```
import plotnine as p9
import pandas as pd
import numpy as np

x = np.random.normal(size=(5000))
y = np.random.normal(size=(5000))

df = pd.DataFrame({"x": x, "y": y})

plt = (p9.ggplot(df, p9.aes(x="x", y="y"))
    + p9.geom_pointdensity())
```

test

Let me know what you think, what I can add, or what can be changed.

Thanks!


has2k1:
I had planned to look at this after the v0.8.0 release, but it has been delayed. So let's see if how far we can get.

Is it possible to add a `count_neighbours` method? And how slow is it and can we optimise it using [numba](https://github.com/numba/numba)?

joshhartmann11:
Sorry this is so late, I just started a new job.  The count method is the same as KDE methods with a uniform kernel.  I implemented a separate method for counting before I realised this.  Let me know if you would still want it implemented.

has2k1:
Yes I want a neighbours method, it more intuitive than densities which many people struggle to understand so this can be a gateway. That it is the same a uniform density is then a bridge into understanding densities. e.g. my thinking when I read that was "really!, ... 10 seconds later ..., of-course it must be".

No worries on time, it isn't critical.

joshhartmann11:
I've added the `kde_count` method which estimates the density as the fraction of samples within some radius.  The performance is comparable to the other KDE methods

test

```
import plotnine as p9
import pandas as pd
import numpy as np

x = np.random.normal(size=(5000))
y = np.random.normal(size=(5000))

df = pd.DataFrame({"x": x, "y": y})

plt = (p9.ggplot(df, p9.aes(x="x", y="y"))
    + p9.geom_pointdensity(package="count"))
```

test2

has2k1:
@joshhartmann11 Thank you so much.

PR #488, opened 2021-02-11
Update index.rst, fix ggplot2 website reference

dsaiztc:
It seems http://ggplot2.org/ is no longer available.


has2k1:
Thank you.

PR #492, opened 2021-02-17
Fix: annotation_logticks + discrete axis + coord_flip & documentation

TyberiusPrime:
This will fail with an AttributeError: 'scale_x_discrete' object has no attribute 'trans'
```python
import plotnine as p9
import plotnine.data
diamonds = p9.data.diamonds.head(100)
g = p9.ggplot(diamonds, p9.aes(x='color', y='price'))
g += p9.geom_point()
g += p9.scale_y_continuous(trans='log10', limits=[10,1000], breaks=[10,100,1000])
g += p9.annotation_logticks('l')
g += p9.coord_flip()           
g
```

When it should end up with logticks at the bottom of the graph.

In addition, the documentation is plain wrong - the sides are *before* the coord-flip,
so that coord-flip does not change the identity of annotated axis.
(We fixed this at one point so that coord-flip is coord-flip, and not coord-flip-but-the-logtick-annotations.
 Apparently  I missed the documentation).

This PR fixes the exception and the documentation and adds the appropriate test cases.


has2k1:
Thanks.

PR #494, opened 2021-02-22
Fix normalization constant for bin counts

ovolve:
Normalization should compute the proportion of elements in each bin relative to the total number of elements, not the number of elements in the biggest bin.

Note: 'ndensity' also looks suspect, but I'm not sure what the intended behavior is there

### DIFF
--- a/plotnine/stats/binning.py
+++ b/plotnine/stats/binning.py
@@ -201,7 +201,7 @@ def result_dataframe(count, x, width, xmin=None, xmax=None):
         'xmax': xmax,
         'width': width,
         'density': density,
-        'ncount': count/np.max(np.abs(count)),
+        'ncount': count/np.sum(np.abs(count)),
         'ndensity': count/np.max(np.abs(density)),
         'ngroup:': np.sum(np.abs(count))})
     return out


has2k1:
The normalisation is such that the counts are scaled so that the maximum count in a group is one. Yes the `ndensity` has an error.

ovolve:
That's a highly non-standard interpretation of a what a histogram shows.

From wikipedia: 

> A histogram may also be normalized to display "relative" frequencies. It then shows the proportion of cases that fall into each of several categories, **with the sum of the heights equaling 1**.

For example seaborn provides the `stat='probability'` option to their `histplot` method to compute this.

But I see that this is following ggplot's convention https://github.com/tidyverse/ggplot2/blob/dbd7d79aa35c49b4eab200cf5f7a084a7748e776/R/bin.R#L180

Given that this is a [frequently asked question regarding ggplot's API](https://www.google.com/search?q=ggplot+normalized+histogram), it would be worth mentioning in the documentation that one is able to do `mapping=aes(y='stat(count) / sum(count)'),` to get the typical view.

has2k1:
I have always found the calculated `density` (whose area is normalized) to be sufficient. i.e it is equivalent to `stat(count/(width*sum(count))`.

PR #500, opened 2021-03-20
Are you interested in a p9 CLI?

ChickenProp:
I have a rudimentary one here. It lets you do e.g.

    # geom point, facet grid
    python -m plotnine --dataset mpg x=displ y=hwy color=class -f grid facets='drv ~ cyl'

    # geom_point + geom_line + stat_smooth
    python -m plotnine --dataset economics x=date y='unemploy/pop' -g point size=0.2 -g line -s smooth method=glm

There's support, at least on some level, for geoms, stats, scales, facets, labels, title. You can take input from a file, stdin, or one of the builtin datasets.

If you're not interested, fair enough. It would be added maintenance burden on you, and I probably won't stick around much to help out with that. It doesn't need to live in this repo, I can keep it separate.

If you are do want it here, I'm happy to take suggestions for what you think it needs before you're happy to merge. Some things it currently lacks:

* For non-builtin datasets, no way to configure csv parsing. A header is required. You can't specify data types or parsing details. (Though you can still do, e.g., `'x=date_col.astype("datetime64")'`.)
* No way to pass an aes or a dataset to a specific layer.
* In general, no way to pass parameters other than strings, ints and floats to anything.
* I think you should be able to use `..foo..` and `stat(foo)` in your aesthetics, but it looks like those are deprecated? There's no way to use `after_stat`.
* No support for themes.
* No documentation.
* No serious thought about how to design the interface.

### DIFF
--- a/plotnine/__main__.py
+++ b/plotnine/__main__.py
@@ -1,0 +1,100 @@
+import sys
+import argparse
+
+import plotnine as p9
+import pandas as pd
+from datetime import date
+
+def parse_kwargs(kwargs):
+    def parse_val(v):
+        try:
+            return int(v)
+        except Exception as e:
+            try:
+                return float(v)
+            except Exception as e:
+                return v
+
+    return { k: parse_val(v) for k,v in [a.split('=', 1) for a in kwargs] }
+
+def parse_scale_kwargs(kwargs):
+    kwargs = parse_kwargs(kwargs)
+    if 'min' in kwargs or 'max' in kwargs:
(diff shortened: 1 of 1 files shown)


has2k1:
@ChickenProp this can certainly be a handy interface, but the best place for it is as a separate package for all the reasons you have pointed out. Plus there would be more room to iterate.

ChickenProp:
Fair enough. I've put it in its own repo: https://github.com/ChickenProp/p9-cli.

PR #507, opened 2021-04-11
Allow user to change relative width or height of facets

tr8dr:
This change allows a user of facet_grid() to specify an optional:

- space={'y': ```<ratios>```}: setting the relative height of vertical facets
  * for example facet_grid("pane ~ .", scales='free_y', space={'y': [2,1,1]}.  This would make the top facet height 2x that of the 2nd and 3rd facet
- space={'x': ```<ratios>```}: setting the relative width of horizontal facets
  * for example space={'x': [1,2,1], ...}
- or a combination of the two:
  * space={'x': [1,2,1], 'y': [2,1,1]}.

A similar extension exists for R ggplot(), though not in the main ggplot package.   Here is an example:
```
import numpy as np
import pandas as pd
import plotnine
from plotnine import *

y = pd.Series(np.random.normal(0.0, 1.0, 400)).cumsum()
df = y.diff().fillna(0.0)
x = np.arange(0,400)

gdf = pd.concat([
    pd.DataFrame({'x': x, 'y': y, 'what': 'y', 'pane': 'cumr'}),
    pd.DataFrame({'x': x, 'y': df, 'what': 'df', 'pane': 'df'}),
])

(ggplot() +
    geom_line(aes(x='x', y='y', color='what'), data=gdf) +
    facet_grid("pane ~ .", scales="free_y", space={'y': [3,1]}))
```
See the attached output included.
example

### DIFF
--- a/plotnine/facets/facet.py
+++ b/plotnine/facets/facet.py
@@ -91,7 +91,7 @@ class facet:
     def __init__(self, scales='fixed', shrink=True,
                  labeller='label_value', as_table=True,
                  drop=True, dir='h',
-                 height_ratios=None, 
+                 height_ratios=None,
                  width_ratios=None):
         from .labelling import as_labeller
         self.shrink = shrink
--- a/plotnine/facets/facet.py
+++ b/plotnine/facets/facet.py
@@ -15,7 +15,7 @@ from ..scales.scales import Scales
 # For default matplotlib backend
 with suppress(ImportError):
     from matplotlib.ticker import locale, FixedFormatter
-	from matplotlib.gridspec import GridSpec
+    from matplotlib.gridspec import GridSpec
 
 
 class facet:
@@ -91,8 +91,8 @@ class facet:
     def __init__(self, scales='fixed', shrink=True,
                  labeller='label_value', as_table=True,
                  drop=True, dir='h',
-				 height_ratios = None, 
-				 width_ratios = None):
+                 height_ratios = None, 
+                 width_ratios = None):
         from .labelling import as_labeller
         self.shrink = shrink
         self.labeller = as_labeller(labeller)
--- a/plotnine/facets/facet.py
+++ b/plotnine/facets/facet.py
@@ -15,6 +15,7 @@ from ..scales.scales import Scales
 # For default matplotlib backend
 with suppress(ImportError):
     from matplotlib.ticker import locale, FixedFormatter
+	from matplotlib.gridspec import GridSpec
 
 
 class facet:
@@ -47,6 +48,10 @@ class facet:
     dir : str in ``['h', 'v']``
         Direction in which to layout the panels. ``h`` for
         horizontal and ``v`` for vertical.
+    height_ratios: for example [2, 1]
+        list of heights (relative ratio) for each vertical row in facet
+    width_ratios: for example [1, 1]
+        list of widths (relative ratio) for each horizontal column in facet
     """
     #: number of columns
     ncol = None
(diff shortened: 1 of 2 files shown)


has2k1:
Great, though I am not sure about the API as compared to  `facet_grid(..., space={'x': [1, 1.5, 1], 'y': [2, 1]})`. The key concern is, currently the `space` parameter is ignored, if it ever is implemented then there would be a clash with `width_ratios` and `height_ratios`.

gokceneraslan:
Btw I have a naive question, if the facet axis variable is categorical and the facet ratios that are nicely implemented here are determined proportionally to the number of categories displayed within each facet, then it's simply `space='free'`, no? Then what is missing for implementing the `space` parameter exactly?

tr8dr:
@has2k1  I did not realize that the `space` parameter could be used in that fashion.  The R ggplot2 docs show free_y, free_x, and fixed as the options for this parameter.   It would be a rather simple change to use the space parameter rather than height_ratios and width_ratios (I think).    Is the intention of:  ```space={'x': [1, 1.5, 1], ...}``` the same in terms of indicating relative ratios?

tr8dr:
@gokceneraslan I am not sure I understand your question above.   I think this implementation could be used as a reference as to how to implement the ```space={'x': [...], 'y': [...]}``` approach.   The code would be (I think):

```
if isinstance(space,dict) and "x" in space:
    xratios = ...
if isinstance(space,dict) and "y" in space:
    yratios = ...

<do the grid arrangement>
```

has2k1:
> Btw I have a naive question, if the facet axis variable is categorical and the facet ratios that are nicely implemented here are determined proportionally to the number of categories displayed within each facet, then it's simply `space='free'`, no? Then what is missing for implementing the `space` parameter exactly?

hmmm. Given a width `w` of the plot panel/facet and say `n` categories, each category will take up **`w/n`** units of space. When you change the width (to say `ω`) of some of the panels, then each category will take **`ω/n`** units of space!

Though, it may be possible (though maybe ugly) to calculate the ratios that correspond to `space='free'`. It would involve probing the limits of each of the scales.

has2k1:
> @has2k1 I did not realize that the `space` parameter could be used in that fashion. The R ggplot2 docs show `free_y`, `free_x`, and `fixed` as the options for this parameter. It would be a rather simple change to use the space parameter rather than height_ratios and width_ratios (I think). Is the intention of: `space={'x': [1, 1.5, 1], ...}` the same in terms of indicating relative ratios?

ggplot2 has those options for the `space` parameter. For plotnine, the width and height ratios control the same aspect of the plot as the `space` parameter, so I think we should add to the `space` parameter.

tr8dr:
@has2k1  I just pushed changes on the branch to utilize the ```space``` parameter rather than the previously separate height_ratios and width_ratios.   I did not address the lack of implementation around other ```space``` options, however made sure that the behavior for those would be consistent with the prior.   You may wish to refactor at some point to unify sizing once or if those other options are activated.

gokceneraslan:
Here is an example of the height ratios proportional to the number of categories, looks very much like `space='free_y'`:

image

plotnine/facets/facet.py
```diff
@@ -47,6 +48,10 @@ class facet:
     dir : str in ``['h', 'v']``
         Direction in which to layout the panels. ``h`` for
         horizontal and ``v`` for vertical.
+    height_ratios: for example [2, 1]
```
gokceneraslan, line 51: Maybe it'd be more consistent to change this also to space?

tr8dr:
I am ambivalent re: whether x/h or y/v;  I think the proposed was originally x and y.  x and y may be easier to remember, however if others prefer h and v, am fine with that.   Could allow both, i.e.:

```space={'y': [2,1]}```. or ```space={'v': [2,1]}```; both would be accepted.

has2k1:
x and y are fine. They are consistent with free_x and free_y.

has2k1:
@tr8dr, I have touched up and merged. Thank you.

PR #517, opened 2021-05-16
fixed truth series error on mapping check

astrocorgi:
bumped into truth series error on input mapping check for plotnine utils. tested and works now.


has2k1:
Thank you for the fix.

PR #521, opened 2021-06-29
Do not iterate over locals()

jsspencer:
This can change the size of locals() during iteration, resulting in a RuntimeError.


has2k1:
Thank you.

PR #525, opened 2021-08-12
fix: annotation_logticks + coord_flip, wrong range used

TyberiusPrime:
annotation_logticks + coord_flip:
  - uses the range of the wrong axis.
  - warns about the wrong axis not being log-transformed.

Minimal example showing both
(I added a debug statement to show the range,
without flip the range used is approximately (-1.5, 9.75):
```
import plotnine as p9
import pandas as pd
import plotnine.data

df = plotnine.data.mtcars.head() # so the x range is not 1..32 but 1..5
g = p9.ggplot(
    pd.concat(
        [
            df,
            df.assign(hp=df.hp * 0.1), # to show the issue is on both sides of the range
            df.assign(hp=df.hp * 1e7), # to show the issue is on both sides of the range
        ],
        axis=0,
    )
)  #
g += p9.geom_point(p9.aes("name", "hp"))
g += p9.scale_y_continuous(trans="log10", name='orig_y')
g += p9.coord_flip()
g += p9.annotation_logticks(sides="l")
g
```

image

This PR:
- makes it use the correct axis, both for warning and for deciding on the range.
- updates the test cases to clearly show of the wrong behavior & verify the ranges

The test cases 'kinda' showed the behavior already, but it was subtle and we never noticed,
this makes it really explicit that something is wrong with the ranges.

The above example now looks like
image


has2k1:
Thank you. This looks right, but the tests (only when running on github-actions) are broken I want to first sort that one out then merge.

has2k1:
@TyberiusPrime, the tests have been fixed. Please rebase onto master.

TyberiusPrime:
> @TyberiusPrime, the tests have been fixed. Please rebase onto master.

It is done.

has2k1:
Thank you.

Issue #177, opened 2018-08-16
Adding subtitle and caption


ghost:
It would be nice to have the caption and subtitle fields. 

From https://ggplot2.tidyverse.org/reference/labs.html

```R
p + labs(title = "New plot title", subtitle = "A subtitle")
```
![image](https://user-images.githubusercontent.com/31109952/44186494-8f23df00-a0ce-11e8-9552-251276b2295d.png)

```R
# The caption appears in the bottom-right, and is often used for
# sources, notes or copyright
p + labs(caption = "(based on data from ...)")
```
![image](https://user-images.githubusercontent.com/31109952/44186498-92b76600-a0ce-11e8-951a-58b96fd19f32.png)

has2k1:
It is related to the same issue (Layout Manager) holding back 46. Trying to implement it without an easy to use Layout Manager would be messy and hard.

raijinspecial:
As an additional feature, would it be possible to obtain regression coefficients like r and r^2 when plotting with stat_ or geom_smooth('lm' or 'ols') and have them print anywhere on the figure? I'm sorry if this is already possible and I have not figured it out.

gresch:
would also appreciate **subtitle** and **caption**. Thank you for this great package.

jaybundy:
Same!

quadrismegistus:
I would also really appreciate this feature. Thanks for plotnine!

gucciwang:
Echoing the above sentiment! :)

himalayajung:
Thank you for all the hard work. Closely following this feature request :)

Ioana-P:
I would also really appreciate this feature; btw Plotnine is amazing, thanks for all your hard work. 

If anyone needs an slightly hacky, interim method for getting a **subtitle** that works **most** of the time, what I've done recently is (for single plots) I've made them into a 1 by 1 facet plot. 
So in pandas I set the subtitle as an additional column, with the same value throughout (e.g., `df['subtitle'] = "Subtitle text"` and passed that as the faceting variable into `facet_wrap.` You can then use the theme params for `strip_text `to make it look as similar to a ggplot sub-title as possible. Naturally this won't work with actual faceted plots, but it's a start. :) 
Haven't figured out anything for captions, yet :/

aeturrell:
Very much looking forward to seeing this feature land. Thanks for the amazing work on the package so far!

has2k1:
Implemented and shipping soon.

Issue #46, opened 2017-08-17
How to draw subplots?


anhqle:
`plotnine` is made using `matplotlib` as the back-end, so I'm guessing there must be a way to draw subplots (without using faceting).

Is there a way to do so? I'd be happy to contribute to the documentation if someone points out a solution.

has2k1:
You cannot draw subplots. It will be implemented when Matplotlib gets a better layout manager. At the moment, it would be hard to implement.

LaDilettante:
Thanks for the quick reply! Is there not even any workaround, such as assigning plotnine plots to variables and print them out in a grid?

has2k1:
There is no work around. Subplot layouts in Matplotlib are not a flexible as they could be, the solution is a [better layout manager](https://github.com/matplotlib/matplotlib/issues/1109).

jklymak:
@has2k1  I've been working on the layout manager, but I'm not clear what flexibility you are looking for.  Since I'm looking for use-cases, can you elaborate (either here or in https://github.com/matplotlib/matplotlib/issues/1109)?  Thanks!

has2k1:
@jklymak, you declared your interest in tacking the layout issue within a day of me finally fiddling with Tillstens examples and trying to figure out what it would take to come up with a workable MEP. 
What I [suggested](https://github.com/matplotlib/matplotlib/issues/1109#issuecomment-319424072) as a plausible way forward has at the minimum (points 1 and 2) the flexibility that would be useful to plotnine.

From the your updates in that thread, I got more assured that the direction you were taking had the main concerns of this project covered. Nonetheless, I intended on working on top of your exploration to make sure of that. That would allow me to raise more substantive concerns. So far I have been reluctant to do that as I think the code you have is still in a lot of flux.

I will create a notebook to track the requirements of this project and test implementations using the constraint layout.

jklymak:
@has2k1  

Sure - you can check out my draft MEP at https://github.com/jklymak/mplconstrainedlayout/blob/master/DraftMEP.ipynb

Its pretty rudimentary. OTOH, I've actually implemented most of it.

gokceneraslan:
Is the layout manager in [matplotlib v2.2](https://github.com/matplotlib/matplotlib/pull/9082) sufficient for proper subplots in plotnine?

has2k1:
[...]

PR #553, opened 2022-01-02
Adds `caption=` option to `labs`

jsoma:
A simple implementation of captions based on the current implementation of plot titles.  I know that in #177 you felt it should wait until matplotlib got its act together, but for my use case I'm required to cite data sources here's hoping.

**Basic usage**

```python
(
    ggplot(df, aes(x='continent'))
    + geom_bar()
    + labs(caption="This is a sample caption")
)
```

**Example output**

caption with x-axis label

caption without x-axis label

caption with default legend

**Customization**

You can also specify the appropriate theme options through `element_text` customize the caption.

```python
(
    ggplot(df, aes(x='continent', color='continent'))
    + geom_bar()
    + labs(caption="This is a sample caption\nacross lines", x="")
    + theme(caption=element_text(
        size=8,
        margin={'r': -120, 't': -30}
    ))
)
```

advanced usage with options, size, and margin


jwhendy:
I just cloned your branch and used this. Very nice! I skimmed the code and can't tell from the comments on #177 and reference to #46 what the pros/cons of  achieving this your way vs. whatever would be preferred re. a Layout Manager (I'm not experienced enough with matplotlib internals)... however from a practical standpoint, this worked great!

jsoma:
I'd say the cons of doing it my way are that it's a bespoke function that isn't terribly elegant, and I'm sure there are edge cases where it doesn't work well. I think "it works with matplotlib as it stands now" is probably the only pro!

jwhendy:
Forgive that I haven't dug in to the existing `labs(title = ...)` or your "bespoke function" as you put it, but is the point that plotnine uses something built-in to matplotlib and your function doesn't? Or that `title` is using something "proper" but you're hacking matplotlib?

I just looked around and noted [`suptitle`](https://stackoverflow.com/questions/1388450/giving-graphs-a-subtitle-in-matplotlib) and [`figtext`](https://stackoverflow.com/questions/34010205/adding-caption-below-x-axis-for-a-scatter-plot-using-matplotlib) as things matplotlib provides... is this general desire for all of titles, subtitles, and captions eventually going to be accomplished through matplotlib mechanisms, or is the point that something entirely new is in the works and being built?

quadrismegistus:
Thank you for this! Being able to cite/caption plots is a must in my field as well so this is a godsend.

has2k1:
@jsoma, thank you for getting started on this. It is not perfect, it can require some finicky fiddling with values, but it is definitely usable.

PR #558, opened 2022-01-20
Update scale_xy.py

zjzh:
refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


TyberiusPrime:
To my mind that's a large loss of readability for no gain (same number of comparisons).

has2k1:
@zjzh, I do not see the improvement.

PR #565, opened 2022-02-14
Test on Python 3.10, drop EOL 3.6

hugovk:
Also upgrade Python syntax with https://github.com/asottile/pyupgrade using `--py37-plus`.


has2k1:
@hugovk, thank you. This was really helpful. I have merged it manually.

PR #568, opened 2022-02-17
Fix `ggplot` docstring

astrojuanlu:


has2k1:
@astrojuanlu, Thank you.

PR #570, opened 2022-03-04
fix: explicitly extract xy coords from 3D-LineString

AlFontal:
I've been dealing with some trajectory calculations which end up producing 3D LineStrings as geometry. 

I am, however, unable to use `geom_map` with this kind of `GeoDataFrame` since `plotnine` expects xy coordinates and fails when the `geometry.coords` object contains tuples with a `z` dimension.

It would also be nice to be able to use the `z` dimension for some other aesthetics, such as color, for instance. My solution for now has been to generate `Point` geometries from the LineStrings and use `geom_path` instead of `geom_map`, but it would be great if this could be handled directly when passing a LineString geometry.

This little fix makes 3D linestrings not throw an error, but does so by completely disregarding the `z` dimension. I am not completely sure of how to fix this to be able to use `z` as an extra aesthetic without explicitly reverting the LineString into Points...

### DIFF
--- a/plotnine/geoms/geom_map.py
+++ b/plotnine/geoms/geom_map.py
@@ -136,7 +136,10 @@ class geom_map(geom):
             segments = []
             for g in data['geometry']:
                 if g.geom_type == 'LineString':
-                    segments.append(g.coords)
+                    if g.has_z:
+                        segments.append([(x, y) for x, y, _ in g.coords])
+                    else:
+                        segments.append(g.coords)
                 else:
                     segments.extend(_g.coords for _g in g.geoms)
 


plotnine/geoms/geom_map.py
```diff
@@ -136,7 +136,10 @@ def draw_panel(self, data, panel_params, coord, ax, **params):
             segments = []
             for g in data['geometry']:
                 if g.geom_type == 'LineString':
-                    segments.append(g.coords)
+                    if g.has_z:
```
has2k1, line 139: Add a warning of type `PlotnineWarning` that the `z` coordinate is being ignored.

has2k1:
There isn't a way to map to the z coordinate. The key issue is, the mapping translates to one aesthetic value for each entry in the dataframe, but for 3-d lines then each entry has a variable number of `z` coordinates!

Though I think with some good thought you may be able create a non-standard custom geom that does that.

plotnine/geoms/geom_map.py
has2k1: Also add test in [test_geom_map.py](https://github.com/has2k1/plotnine/blob/15b9759047a46523d9c5a525f6f60dac24b6b72a/plotnine/tests/test_geom_map.py).

PR #571, opened 2022-03-04
Updated an instance of df.append() called out by latest version of pandas

tabedzki:
I updated the pandas and got the following deprecation warning:

${anaconda3_home}/lib/python3.9/site-packages/plotnine/utils.py:371: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead

Naively implemented an update without validating. I don't know how to use a local package to test.

### DIFF
--- a/plotnine/utils.py
+++ b/plotnine/utils.py
@@ -368,7 +368,7 @@ def join_keys(x, y, by=None):
     if isinstance(by, tuple):
         by = list(by)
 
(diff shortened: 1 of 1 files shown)


has2k1:
Thanks. This has been merged in https://github.com/has2k1/plotnine/commit/e9cc70dd544ff90e71b4de7360006315e4dba989 .

Issue #575, opened 2022-03-13
xlim and ylim on qplot have no effect


Gaasmann:
Hello,

I was testing qplot() and it seems the parameters xlim and ylim have no effect. I can't find them used in the qplot function's code either.

An example:
```python
from plotnine import *
from plotnine.data import economics
qplot("carat", data=diamonds, facets="color ~ .", geom="histogram", binwidth=0.1, xlim=(0,3))
```
![image](https://user-images.githubusercontent.com/13124500/158076234-61f186e3-fceb-4354-b95a-a0e3ded78bde.png)

I'm using plotnine 0.8.0 from PyPI with Python 3.10.

Thanks,

-- 
Nicolas

has2k1:
Fixed in #577.

PR #577, opened 2022-03-28
Fix xlim and ylim effect for qplot

joshhartmann11:
Fixes issue #575

From the issue, the following code:
```
from plotnine import *
from plotnine.data import diamonds
p = qplot("carat", data=diamonds, facets="color ~ .", geom="histogram", binwidth=0.1, xlim=(0,3))
```
Now produces the plot:
test_diamond


plotnine/qplot.py
```diff
@@ -15,6 +15,8 @@
 from .utils import Registry, is_string, array_kind
 from .exceptions import PlotnineError, PlotnineWarning
 from .scales import scale_x_log10, scale_y_log10
+from .scales import xlim as scales_xlim
+from .scales import ylim as scales_ylim
```
joshhartmann11, line 19: Thoughts on a better solution for this naming conflict?

has2k1: Use the general `lims` instead of the specialised functions.

has2k1:
Thank you

PR #584, opened 2022-04-21
Adding log-transformed density option for fill in stat_bin_2d

bmwil3:
Minor change to add the ability to fill tiles in stat_bin_2d with the log10-transformed density for skewed joint distributions.


has2k1:
Sorry, I forgot to comment & you have probably figured it out. You can map to the log-density with `after_stat('np.log10(density)')`.

PR #592, opened 2022-05-20
Add a space in CI warning

nolanbconaway:
I saw this warn and just _had_ to toss a 2 character PR your way! 

```
PlotnineWarning: Confidence intervals are not yet implementedfor lowess smoothings.
```


has2k1:
Thank you. Can you just omit the `+` and add a space at the end of the first string.

Issue #582, opened 2022-04-19
Enabling 3D plots?


krassowski:
Could 3D plots be supported by plotnine, or if not, could the internals be slightly adjusted to enable easy extension to 3D plots? Currently with a simple hack it is possible to get a simple 3d plotnine plot:

```python
from plotnine import scale_color_manual, scale_shape_manual, theme_minimal
from plotnine.data import mtcars

mtcars['transmission'] = mtcars['am'].replace({0: 'automatic', 1: 'manual'})
(
    ggplot_3d(mtcars)
    + aes(x='hp', y='disp', z='mpg', shape='transmission', color='transmission')
    + geom_point_3d()
    + theme_minimal()
    + scale_shape_manual(values={'automatic': 'o', 'manual': '*'})
    + scale_color_manual(values={'automatic': 'red', 'manual': 'blue'})
)
```

![image](https://user-images.githubusercontent.com/5832902/163988567-c27f6295-e61a-4b51-8359-b2ede4e9b878.png)

The hack is:

```python
from plotnine import ggplot, geom_point
from plotnine.utils import to_rgba, SIZE_FACTOR


class ggplot_3d(ggplot):
    def _create_figure(self):
        figure = plt.figure()
        axs = [plt.axes(projection='3d')]
        figure._themeable = {}
        self.figure = figure
        self.axs = axs
        return figure, axs

    def _draw_labels(self):
        ax = self.axs[0]
        ax.set_xlabel(self.layout.xlabel(self.labels))
        ax.set_ylabel(self.layout.ylabel(self.labels))
        ax.set_zlabel(self.labels['z'])


class geom_point_3d(geom_point):
    REQUIRED_AES = {'x', 'y', 'z'}

    
    @staticmethod
    def draw_unit(data, panel_params, coord, ax, **params):
        size = ((data['size']+data['stroke'])**2)*np.pi
        stroke = data['stroke'] * SIZE_FACTOR
        color = to_rgba(data['color'], data['alpha'])

        if all(c is None for c in data['fill']):
            fill = color
        else:
            fill = to_rgba(data['fill'], data['alpha'])

        ax.scatter3D(
            data['x'],
            data['y'],
            data['z'],
            s=size,
            facecolor=fill,
            edgecolor=color,
            marker=data.loc[0, 'shape'],
        )

    @staticmethod
    def draw_group(data, panel_params, coord, ax, **params):
        data = coord.transform(data, panel_params)
        units = 'shape'
[...]

PR #606, opened 2022-07-24
Enable 3D subplots support for facets

krassowski:
Enable 3D subplots support for facets when custom layout with 3d projection is provided, partially addressing #582.

There are three changes:
- adding `projection` attribute to `Layout`
- allowing to substitute `Layout` in `ggplot` subclasses
- converting `FancyBboxPatch` to 3D if 3D axes are detected (without it all `facet_*()` would need be subclassed only to update the `strip()` call with a custom subclass of it).

I wonder if `projection` attribute could be also useful for `coord_polar` as it can be set to "polar"; if that is the case and therefore it should be moved elsewhere (other than layout), I would be happy with that as well.

Example of what it enables:

image

### DIFF
--- a/plotnine/facets/facet.py
+++ b/plotnine/facets/facet.py
@@ -317,7 +317,7 @@ class facet:
         """
         Create suplots and return axs
         """
-        num_panels = len(layout)
+        num_panels = len(layout.layout)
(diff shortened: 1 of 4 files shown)


has2k1:
This PR stalled and does not address a major issue. Future work will most definitely address this issue.

PR #607, opened 2022-07-24
Make `geom_text` and `geom_point` easier to subclass

krassowski:
Part two for #582:

- allows to extend `draw_unit` of `geom_text` without the need to redefine `draw_group` in a subclass
- allows to add extra attributes in subclasses of `geom_text` without the need to re-define `draw_group` (by providing custom params/data columns mapping in class attributes); it is likely a little bit faster as columns are copied at once in the `.rename()` call, rather than one-column-at-a-time as before.

### DIFF
--- a/plotnine/geoms/geom_point.py
+++ b/plotnine/geoms/geom_point.py
@@ -31,14 +31,13 @@ class geom_point(geom):
         """
         self.draw_group(data, panel_params, coord, ax, **params)
 
-    @staticmethod
-    def draw_group(data, panel_params, coord, ax, **params):
+    @classmethod
+    def draw_group(cls, data, panel_params, coord, ax, **params):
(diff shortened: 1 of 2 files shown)


plotnine/geoms/geom_point.py
```diff
@@ -31,14 +31,13 @@ def draw_panel(self, data, panel_params, coord, ax, **params):
         """
         self.draw_group(data, panel_params, coord, ax, **params)
 
-    @staticmethod
-    def draw_group(data, panel_params, coord, ax, **params):
+    @classmethod
```
has2k1, line 34: For uniformity and since it is backwards compatible, change all `geom.draw_group` to classmethod.

has2k1:
This PR stalled and does not address a pressing issue. Future work will most definitely address this issue.

PR #615, opened 2022-08-30
add subtitle arguments into labels.py

Anggi-Permana-Harianja:
Adding subtitle arguments into labels.py which is currently not supported.

Example:

labs(x = "GDP Per Capita", y = "Life Expectancy in Years",
         title = "Economic Growth and Life Expectancy",
         subtitle = "Data points are country-years",
         caption = "Source: Gapminder.")

will throw an error because the current version has no subtitle arguments. Having a subtitle in the plot has been around for a long time in R

### DIFF
--- a/plotnine/labels.py
+++ b/plotnine/labels.py
@@ -3,8 +3,8 @@ from copy import deepcopy
 from .mapping.aes import rename_aesthetics, SCALED_AESTHETICS
 from .exceptions import PlotnineError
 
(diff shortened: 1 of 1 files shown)


has2k1:
What you have is not enough to get subtitles working. As it stands, subtitles are not easy to implement well. You have to make space for them between the plot and the title the current layout mechanism cannot handle that well. This is the bigger issue, that we hope to tackle.

PR #636, opened 2022-10-18
fix passing of dparams when using stat_qq

### LINKED ISSUE #635: `stat_qq` doesn't accept `dparams`
Coming from [this SO post](https://stackoverflow.com/questions/70798622/qq-plot-in-python-using-plotnine#70798622), `stat_qq` gives an error when you try to pass a dict as arguments to `dparams`:

```
from plotnine import *
import numpy as np

n = 207
values = -1 + np.random.beta(n/2-1, n/2-1, 100) * 2 # my data

params = {'a':n/2-1, 'b':n/2-1, 'loc':-1, 'scale':2}
ggplot(aes(sample = values)) + stat_qq(distribution = 'beta', dparams = params)
# gives TypeError: '>' not supported between instances of 'numpy.ndarray' and 'int'
```

I think this is due to [`compute_group`](https://github.com/has2k1/plotnine/blob/main/plotnine/stats/stat_qq.py#L81) of class `stat_qq` unpacking `dparams` with `*` instead of `**` for kwargs, as `dparams` expects a dict.

jonas-hag:
`stat_qq` unpacks `dparams` in wrong way when passing them to `cdist.ppf` resulting in an error so far; this fixes #635 

I hope the PR is ok, I wasn't sure if/how I should add a test, I'm happy about feedback.

### DIFF
--- a/plotnine/stats/stat_qq.py
+++ b/plotnine/stats/stat_qq.py
@@ -78,6 +78,6 @@ class stat_qq(stat):
 
         quantiles = np.asarray(quantiles)
         cdist = get_continuous_distribution(params['distribution'])
-        theoretical = cdist.ppf(quantiles, *params['dparams'])
+        theoretical = cdist.ppf(quantiles, **params['dparams'])
         return pd.DataFrame({'sample': sample,
                              'theoretical': theoretical})
--- a/plotnine/stats/stat_qq.py
+++ b/plotnine/stats/stat_qq.py
@@ -60,7 +60,7 @@ class stat_qq(stat):
     DEFAULT_AES = {'x': after_stat('theoretical'), 'y': after_stat('sample')}
     DEFAULT_PARAMS = {'geom': 'qq', 'position': 'identity',
                       'na_rm': False,
-                      'distribution': 'norm', 'dparams': (),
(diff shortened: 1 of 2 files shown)


plotnine/stats/stat_qq.py
```diff
@@ -78,6 +78,6 @@ def compute_group(cls, data, scales, **params):
 
         quantiles = np.asarray(quantiles)
         cdist = get_continuous_distribution(params['distribution'])
-        theoretical = cdist.ppf(quantiles, *params['dparams'])
+        theoretical = cdist.ppf(quantiles, **params['dparams'])
```
has2k1, line 81: You need to change the [default](https://github.com/has2k1/plotnine/pull/636/files#diff-5374bfbe62ced2d754f62c16a97624db7b812869ab590bf1a230b521046a7162R63) value for `dparams` to be a `{}`.

jonas-hag: thanks you, I've changed it

jonas-hag:
I've also changed the default to dict for stat_qq_line so that it is the same as for stat_qq

has2k1:
@jonas-hag, thank you.

PR #644, opened 2022-11-05
Add minimal typecheck infrastructure to CI

Emerentius:
I've added a mypy.ini file to configure the mypy typechecker and added a CI job to run mypy during testing.
The config is set such that you can't forget to type all functions in a typed module. A few guard rails against untyped `Any` values sneaking in are also enabled. Errors are deactivated unless opted-in for a specific module.
I've done just that for 3 modules as an example and to test it.

I saw two errors that seem to be bugs.
1. `plotnine.exceptions.warning_format` accepted an additional default argument that wasn't the documentation of `warnings.formatwarning` (harmless probably)
2. `plotnine.options.set_option` and `get_option` are attempting to importing the exceptions module from one level too high.


plotnine/labels.py
```diff
@@ -17,17 +22,17 @@ class labs:
         Aesthetics (with scales) to be renamed. You can also
         set the ``title`` and ``caption``.
     """
-    labels = {}
+    labels: dict[str, str] = {}
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs: dict[str, str]) -> None:
```
has2k1, line 27:
```python
def __init__(self, **kwargs: str) -> None
```

Ref: https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated

mypy.ini
```diff
@@ -0,0 +1,25 @@
...
+# The set of typechcecked modules can be incrementally expanded until everything
+# is correctly typed.
+warn_return_any = True
+warn_unused_configs = True
+disallow_untyped_defs = True
+disallow_any_generics = True
+disallow_any_decorated = True
+allow_redefinition = True
+warn_unreachable = True
+ignore_errors = True
```
has2k1, line 15: Include the description of each option, e.g.


```ini
# Allows variables to be redefined with an arbitrary type, as long as the
# redefinition is in the same block and nesting level as the original definition.
allow_redefinition = True
```

has2k1 (changes requested):
@Emerentius, thank you. This is a sensible roadmap to jumpstart type checking.

plotnine/labels.py
Emerentius: Right, the dict[str, _] part is implicit.

mypy.ini
Emerentius: I've added the descriptions. I've also removed `disallow_any_decorated`. The problem with that setting is, that it doesn't just warn if you have a typed definition that loses its type hints because of the decorator, it also warns if your original definition contains Any and has a decorator.

That means this pattern that appears in the codebase a lot generates warnings.
```
@staticmethod
def foo(a: str, b: int, **kwargs: Any): # kwargs is forwarded to some other function
    ...
```

has2k1:
Thanks a bunch.

PR #648, opened 2022-11-10
Type hint user-facing methods of ggplot (and watermark)

Emerentius:
Add type hints to the public facing API of the central ggplot class (plus the watermark class). Some other functions are also type hinted to aid in the typing of the ggplot class.
I've added a submodule `typing` to add protocols and unions of general concepts that apply across the library. Numpy uses a submodule of the same name. Should I make this module pseudo-private (e.g. rename it to `_typing`) in the beginning?
Eventually, it should be public.

Mypy has released a new version 2 days ago and implicit optional types have been removed. I've addded explicit notations to fix the errors. Error codes are now shown by default so I've removed the setting from mypy.ini.

A note on import style, I see you are preferring relative imports for intra-package imports and I'm trying tried to conform to this, but there are cases where a relative import will result in a type error because mypy thinks I'm referring to the module and not the class. That's where I'm falling back to absolute imports. It happens in cases like this
```python
from . import ggplot
foo: ggplot  # err about a module not being valid as type
```

### DIFF
--- a/plotnine/ggplot.py
+++ b/plotnine/ggplot.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import sys
 from collections.abc import Sequence
 from copy import deepcopy
@@ -8,6 +9,7 @@ from warnings import warn
 
 import pandas as pd
 import matplotlib as mpl
+import matplotlib.figure
 import matplotlib.pyplot as plt
 import matplotlib.transforms as mtransforms
 from matplotlib.offsetbox import AnchoredOffsetbox
@@ -24,7 +26,11 @@ from .exceptions import PlotnineError, PlotnineWarning
 from .scales.scales import Scales
 from .coords import coord_cartesian
 from .guides.guides import guides
-from .geoms import geom_blank
+
+# mypy believes there is a duplicate definition
+# of geom_blank even though it only appears once
+from .geoms import geom_blank   # type: ignore[no-redef]  # mypy bug
+
 from .utils import (
     defaults,
     from_inches,
@@ -34,6 +40,12 @@ from .utils import (
     ungroup
 )
 
+import typing
+from typing import Any, Iterable, Union
+
+if typing.TYPE_CHECKING:
+    import plotnine as p9
+    from .typing import DataLike, PlotAddable
 
 # Show plots if in interactive mode
 if sys.flags.interactive:
@@ -59,7 +71,12 @@ class ggplot:
         in which `ggplot()` is called.
     """
 
-    def __init__(self, data=None, mapping=None, environment=None):
+    def __init__(
+        self,
+        data: DataLike | None = None,
+        mapping: aes | None = None,
+        environment: dict[str, Any] | None = None
+    ) -> None:
         # Allow some sloppiness
         data, mapping = order_as_data_mapping(data, mapping)
         if mapping is None:
@@ -68,7 +85,7 @@ class ggplot:
         # Recognize plydata groups
         if hasattr(data, 'group_indices') and 'group' not in mapping:
             mapping = mapping.copy()
-            mapping['group'] = data.group_indices()
+            mapping['group'] = data.group_indices()  # type: ignore
 
         self.data = data
         self.mapping = mapping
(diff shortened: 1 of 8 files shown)


plotnine/utils.py
```diff
@@ -1007,9 +1034,9 @@ def order_as_data_mapping(*args):
     elif n == 1:
         single_arg = ungroup(args[0])
         if isinstance(single_arg, pd.DataFrame):
-            return None, single_arg
-        elif isinstance(single_arg, aes):
             return single_arg, None
+        elif isinstance(single_arg, aes):
+            return None, single_arg
         else:
```
Emerentius, line 1040: The order of the return values here was flipped from the 2-arg-case. As far as I can see, everything but the 2 argument case is dead code, so that's why it never came up.

plotnine/typing.py
```diff
@@ -0,0 +1,43 @@
...
+    import plotnine as p9
+
+    # Note that dunder methods may have special-casing in mypy's handling
+    # where it checks for consistency between different functions or
+    # enforces the absence of additional arguments.
+    # In the case of __radd__, the inplace argument will cause errors
+    # but having the protocol here still lets it typecheck the return
+    # value correctly.
+    class PlotAddableInPlace(Protocol):
+        def __radd__(self, other: p9.ggplot, inplace: bool = ...) -> p9.ggplot:
```
has2k1, line 17: `inplace` parameter has been removed.

plotnine/utils.py
```diff
@@ -1041,7 +1068,10 @@ def order_as_data_mapping(*args):
     return data, mapping
 
 
-def is_data_like(obj):
+# Returning a type guard here is not fully sound, because if `obj`
+# is a callable, we aren't checking that it has no required args
+# and we can't check the return value's type.
+def is_data_like(obj: Any) -> TypeGuard[DataLike]:
```
has2k1, line 1074: This is a little head spinner. At the moment I have no idea how to clear it up nicely. We can inspect, but that is overkill. Maybe remove it later!

has2k1 (changes requested):
I have done some refactoring to remedy to the highlights in comments. They have been very helpful as they allowed me to make changes for the better.

plotnine/utils.py
Emerentius: The type information is available in the user's code when they construct any of the various classes that accept data, it just gets lost along the way because both data and aesthetic mappings are accepted as positional params in any order and are then rearranged to fit.

The code could be rearranged so the type information is never lost.

plotnine/ggplot.py
```diff
@@ -34,6 +39,12 @@
     ungroup
 )
 
+import typing
+from typing import Any, Iterable, Union, overload
```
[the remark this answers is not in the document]
Emerentius: Is the order in which isort puts the imports acceptable? Then I'd just run that on any file I modify. It may reorder a few of the other imports as well.

plotnine/utils.py
Emerentius: The new version wouldn't pass the typechecker, because it couldn't prove that it would always return something. And in fact, if there were two mappings or two data arguments, then the code would just run through all the checks, never fail and return None. I've rewritten it so that can't happen.

Emerentius:
Now that `to_pandas()` has been removed, the conversion will need to be done at some later point. There is only one place that does the conversion and that is `layer._make_layer_data()`. Is that enough?
I had to modify that function a bit to get it to typecheck for the case where `layer.data` is a `Callable[[DataFrame], DataFrame]` and the input `plot_data` hasn't been converted to a DataFrame yet.

plotnine/ggplot.py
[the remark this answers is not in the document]
Emerentius: I changed the import order. `isort` modified a lot more than I wanted to commit so I ended up not using it.

plotnine/utils.py
Emerentius: I've narrowed the input type from `Any` to `DataLike | aes | None`. In that case, if the input is a Callable, then it must have the right signature or a type error would be thrown in the user's code, if they use a type checker.

Even with that, there is still the problem that the different allowed input types like callables and objects with `to_pandas` methods are not mutually exclusive. If an object has a `to_pandas()` method with the right signature but is also callable with the wrong signature, then the type checker would accept it. When conversion is then attempted by calling the object first, that would fail and a runtime error would be thrown. The inverse problem could also happen if `to_pandas()` conversion were attempted first in the library.

There are a couple options:
1. Require the user to specify the conversion method
2. Try the other conversion methods, if one fails
3. Do nothing of the sort and just live with the corner case because objects that are both callable and have a to_pandas() method aren't common (no idea if this is true)

has2k1: Option 3, "Do nothing" is enough. To my knowledge `df.to_pandas()` only applies to polars objects. Also, the future is probably polars with first class support then pandas frames will be converted, but that is still further down the line.

has2k1:
I am happy with this. I will look into `isort`, I think it now has enough options.

has2k1:
Thanks @Emerentius, I have learnt a lot.

PR #650, opened 2022-12-01
legend: use figure artist instead of axis artist

stettberger:
When positioning a legend in an faceted plot in an facet != ax[0], the legend is not drawn in front of the facets, but behind them. This behavior is different from ggplot.

Without the Patch:

broken

With the Patch:

good


stettberger:
As far as I understand the code coverage report, the checker complains that I've removed a line?

has2k1:
Thank you, this is was a correct solution. I have since implemented a layout manager that allows for better positioning of the legend and predictable image sizes. And there will be more related improvements.

PR #651, opened 2022-12-01
geom_text: do not eliminate NaN when format_string or parse is given

stettberger:
As string.format happily replaces an NaN with "nan", we forget the information that a value was missing, whereby it is no longer removed by na_rm, but the string "nan" is displayed.


has2k1:
@stettberger, thank you. I improved the solution to work for non float columns.

Issue #390, opened 2020-04-26
Idea: Experiment with matplotlib's tight_layout


has2k1:
`tight_layout` autospaces axes and their adornments. Maybe we could make some types of multi-plots possible if we could attach artists to axes and they would be considered during tight layout computation. This requires submitting a PR to Matplotlib.

[This](https://gist.github.com/thesamovar/52dbbb3a58a73c590d54c34f5f719bac) may help in the development.

Related: #46

PR #662, opened 2023-02-15
Save file with actual pixel dimensions

wch:
When a figure has a specified size, the tight layout causes the outside to be trimmed off, so the resulting .png file is smaller than specified. For example, one might expect this code to generate a 600x400 image.

```py
from plotnine import *

p = ggplot(mtcars, aes('wt', 'mpg')) + geom_point()
p.save("test.png", width=6, height=4, dpi=100)
```

The result on my Mac, however, is 528x374. This is a result of using `fig.savefig(bbox_inches='tight')`. Here is the generated png:

test-orig

With this PR, the resulting image is the exact specified dimension. It does this by calling `fig.tight_layout()` before `fig.savefig()`. Here is the generated png after the change:

test

### DIFF
--- a/plotnine/ggplot.py
+++ b/plotnine/ggplot.py
@@ -659,7 +659,6 @@ class ggplot:
         Use it to get access to the figure that will be saved.
         """
         fig_kwargs: Dict[str, Any] = {
-            "bbox_inches": "tight",  # 'tight' is a good default
             "format": format,
         }
         fig_kwargs.update(kwargs)
@@ -774,6 +773,7 @@ class ggplot:
             verbose=verbose,
             **kwargs,
         )
+        sv.figure.tight_layout()
         sv.figure.savefig(**sv.kwargs)
 
 


has2k1:
I have to think about how to go about this one.

Here is some context on `plt.tight_layout()`.

Initially it was marked experimental for a long time, and after that it still could not deal with some artists that we place around the panels. Occasionally it could surprise with some layout arrangements.

But it has had some work in the last year and it's API has been spelt out better. I think the issues I had related to this one may also been been resolved. If so, the solution is to refactor the layout arrangement in plotnine to feed spacing parameters into `tight_layout`, Or to define our own layout_engine that is based on tight_layout.

While some of the larger layout goals may not be possible, I think this time it will be possible to get fixed size figures with layouts :). If done, this will be my happiest tick off in a long time. 

Ref:
https://github.com/has2k1/plotnine/issues/390
https://github.com/matplotlib/matplotlib/pull/4477
https://github.com/matplotlib/matplotlib/pull/4466
[layout_engine](https://matplotlib.org/stable/api/layout_engine_api.html#module-matplotlib.layout_engine) (new in MPL-3.6, Sep'22)

PR #674, opened 2023-03-24
fix potential AttributeError

yunfeiguo:
AttributeError: 'numpy.ndarray' object has no attribute 'isnull' for some input


has2k1:
I do not see how the error comes about. I think the object is guaranteed to be pandas series at that point.

PR #678, opened 2023-04-18
Fix typo in ggpy URL in about page

ressy:
The [About plotnine](https://plotnine.readthedocs.io/en/stable/about-plotnine.html) page has a backtick at the end of the ggpy URL.  This would remove that character to fix the link.


has2k1:
Thank you.

PR #679, opened 2023-04-18
TYP: Fix type hints in animation

ksunden:
Actual return is in fact list[list[artist]], which is what mpl.ArtistAnimation actually expects, just the type hints were incorrect

Test case (adapted from tests):

```python
from plotnine.animation import PlotnineAnimation
from plotnine import labs, lims, qplot, theme_minimal, theme
import matplotlib.pyplot as plt

_theme = theme(subplots_adjust={"right": 0.8})

x = [1, 2, 3, 4, 5]
y = [1, 2, 3, 4, 5]
colors = [[1, 2, 3, 4, 5], [2, 3, 4, 5, 6], [3, 4, 5, 6, 7]]

def plot(i):
    return (
        qplot(x, y, color=colors[i], xlab="x", ylab="y")
        + lims(color=(1, 7))
        + labs(color="color")
        + theme_minimal()
        + _theme
    )

plots = [plot(i) for i in range(3)]
anim = PlotnineAnimation(plots, interval=100, repeat_delay=500)

fig, artists = anim._draw_plots(plots)
print(artists)
print(type(artists[0]))
```

Outputs (formatted for readability):
```
[[<matplotlib.collections.PathCollection object at ...>, <matplotlib.offsetbox.AnchoredOffsetbox object at ...>],
 [<matplotlib.collections.PathCollection object at ...>, <matplotlib.offsetbox.AnchoredOffsetbox object at ...>],
 [<matplotlib.collections.PathCollection object at ...>, <matplotlib.offsetbox.AnchoredOffsetbox object at ...>]]
<class 'list'>
```

Note that `_draw_plots` returns a nested list.

We (matplotlib) recently introduced type hints (matplotlib/matplotlib#24976), this was discovered in my survey of some of our downstream packages to see how our type hints do.
There are a couple more things that I have flagged that either I have opened or intend to open fixing things on our end, I think there may be some clarifications for the type checker here still even so (though mostly fairly easy e.g. "add a type hint for this variable which the typechecker just sees as Unknown" or "make this list a tuple so the typechecker can validate length and each element type")
We would invite you to test it out early so that you are not surprised when mpl v3.8 is released in a couple months.
If you have any questions or are unsure of how to satisfy the typechecker for mpl related things, feel free to ping me.

### DIFF
--- a/plotnine/animation.py
+++ b/plotnine/animation.py
@@ -73,13 +73,13 @@ class PlotnineAnimation(ArtistAnimation):
 
     def _draw_plots(
         self, plots: Iterable[Ggplot]
-    ) -> tuple[Figure, list[Artist]]:
+    ) -> tuple[Figure, list[list[Artist]]]:
         with pd.option_context("mode.chained_assignment", None):
             return self.__draw_plots(plots)
 
     def __draw_plots(
         self, plots: Iterable[Ggplot]
-    ) -> tuple[Figure, list[Artist]]:
+    ) -> tuple[Figure, list[list[Artist]]]:
         """
         Plot and return the figure and artists
 


has2k1:
Thank you for the heads up and PR. Type hinting in matplotlib is quite welcome and I will keep track ahead of the 3.8 release.

PR #690, opened 2023-06-05
ENH: Add BibTeX citation file

schackartk:
I would like to start off with mentioning that I really enjoy using plotnine!

To make citing it easier, I am opening this pull request adding a BibTeX citation file. Such a file can be easily imported to reference managers such as Zotero.

If you are interested in including such a file, I am of course happy to make any changes the developers would like. A few considerations are:

- Would you prefer a more detailed author list?
- For the URL, is the Zenodo address preferred?
- I used the DOI that will always point to the latest version in Zenodo. This is to simplify maintenance of this file. However, it is possible to keep the DOI up to date with each release.
- Similar to the above point, the date of each release could be added and updated.

Thank you for your consideration of this pull request.

-Ken

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,0 +1,6 @@
+@misc{plotnine,
+  author = {The plotnine development team},
+  title  = {plotnine: A grammar of graphics for Python},
+  url    = {https://github.com/has2k1/plotnine},
+  doi    = {https://doi.org/10.5281/zenodo.1325308}
+}


has2k1:
Other examples
https://github.com/scipy/scipy/blob/404d3df57ce63581b076077a04a918d880d4f862/CITATION.bib
https://github.com/matplotlib/matplotlib/blob/4f65e66d1dcf9788d542fc893ee2e40940304da1/CITATION.bib

Issue #653, opened 2022-12-12
Feature request: ability to add borders to colorbar


AlFontal:
### Discussed in https://github.com/has2k1/plotnine/discussions/652

<div type='discussions-op-text'>

<sup>Originally posted by **AlFontal** December 12, 2022</sup>
I've been trying to add borders to colorbars to be consistent with other colorbars generated in `matplotlib`, which by default contain black borders.

I could always remove the borders on the `matplotlib` side, but I've come to quite like the aesthetics of a bordered colorbar, and I was wondering whether it's even possible to replicate that.

As an extra, `ggplot2`'s current release allows to do this by using the `frame.colour` attribute in `guide_colorbar`.

Does anyone know whether there is any way of doing this through the current `plotnine` API?

Minimal example of what I mean:

```python
df = pd.DataFrame({
     'value': np.random.randn(100),
     'variable': np.tile(['A', 'B', 'C', 'D', 'E'], 20),
     'time': np.repeat(np.arange(1, 21), 5)
})
```

```python
# in matplotlib
plt.imshow(df.pivot('time', 'variable', 'value'))
plt.colorbar()
```

<img width="300" alt="imagen" src="https://user-images.githubusercontent.com/10958332/207078375-24181036-0e73-4dca-a0cc-5a3e9ff53380.png">

```R
# ggplot2 in R
ggplot(df) +
    aes(variable, time, fill = value) +
    geom_tile() +
    scale_fill_continuous(type = 'viridis',
                          guide = guide_colorbar(frame.colour = 'black'))
```
![imagen](https://user-images.githubusercontent.com/10958332/207080391-01ba89e3-78ef-4d19-b02a-8e6cddbdac03.png)

</div>

has2k1:
This has been completed, but it only allows the newer API where style can be controlled through the `theme` parameter or the global `theme`.

PR #701, opened 2023-07-12
ENH: support more style params for `guide_colorbar`

daylinmorgan:
This PR addresses #653, along with adding the other related parameters from the `ggplot2` implementation.

Example:
```python
import numpy as np
import pandas as pd

from plotnine import (
    aes,
    element_blank,
    geom_tile,
    ggplot,
    guide_colorbar,
    scale_fill_continuous,
    theme,
)

df = pd.DataFrame(
    {
        "value": np.random.randn(100),
        "variable": np.tile(["A", "B", "C", "D", "E"], 20),
        "time": np.repeat(np.arange(1, 21), 5),
    }
)

(
    ggplot(df, aes("variable", "time", fill="value"))
    + geom_tile()
    + scale_fill_continuous(
        guide=guide_colorbar(frame_color="black", ticks_color="black")
    )
    + theme(panel_background=element_blank(),figure_size=(3,3))
)
```

colorbar

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -10,9 +10,12 @@ v0.12.2
 Enhancements
 ************
 
+- :class:`~plotnin.guide.guide_colorbar` Now supports additional 
+  styling parameters: `frame_color`, `frame_linewidth`, 
+  `ticks_color`, `ticks_linewidth`.
+
 - All `__all__` variables are explicitly assigned to help static typecheckers
   infer module attributes. (:issue:`685`)
-
 
 v0.12.1
 -------
--- a/plotnine/guides/guide_colorbar.py
+++ b/plotnine/guides/guide_colorbar.py
@@ -33,8 +33,16 @@ class guide_colorbar(guide):
         a smoother colorbar. Default is 20.
     raster : bool
         Whether to render the colorbar as a raster object.
+    frame_color : None
+        Color of surrounding frame. If None, no frame is drawn.
+    frame_linewidth : float
+        Line width of frame. Default is 1.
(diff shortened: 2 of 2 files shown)


has2k1:
@daylinmorgan I will look at this before of the next release. The styling needs to be integrated with the theming system. You can take a crack at it if you wish, otherwise I will do it.

daylinmorgan:
>  The styling needs to be integrated with the theming system. 

Could you elaborate more what you mean by this? Do these overlap with any of the current themeable elements?

has2k1:
> Could you elaborate more what you mean by this?

The rectangle for the colorbar should be themeable with `element_rect(...)` and, the lines for the ticks should be themeable with `element_line(...)`. 

How it works is, we store the Matplotlib artists that should be themeable in a dictionary `theme._targets`. Then a class in `themes/themeable.py` does the actual "theming" i.e. setting the properties of the artist.

daylinmorgan:
Ah! It wasn't clear just using the docs for `ggplot2` that the [`guide_colorbar`](https://github.com/tidyverse/ggplot2/blob/bde88f82a1c3343f7172a38aa4f3e4cef34777f3/R/guide-colorbar.R#L139-L152C1) actually supports a theme object as well.

has2k1:
I have merged this in https://github.com/has2k1/plotnine/commit/18f397e4 and completely refactored the guides to mainly be styled through the local `theme` parameter or the global theme. Plus there is now full control over the positioning of guides.

daylinmorgan:
Sounds great! Looking forward to the next release.

PR #736, opened 2023-12-21
FIPS Compliance

ccravens:
Currently plotnine uses `md5` which is disabled in FIPS-compliant environments:

image

I've updated `md5` -> `sha256` to start the process of making the plotnine library FIPS-compliant.


has2k1:
@ccravens, thanks. I had never heard of FIPS complaince.

PR #793, opened 2024-05-29
Update README.md

EvaMaeRey:
small suggestion.  I think new wording would speak to aesthetic mapping being separate declaration from layer declaration.  I think new wording is still high-level and accessible to newcomers.


has2k1:
Thank you @EvaMaeRey, the clarity `>>` size of the change.

PR #799, opened 2024-06-13
Added argument in show() to speficy figure number to use

fkgruber:
I added a feature allow to select the figure number to use to plot. This way you can work simultaneously with 2 different plot displayed in 2 different figure windows.

Example usage
```

import plotnine as pn
import pandas as pd

ggp1=(
     pn.ggplot(pd.DataFrame(
         {"x":[1,2,3,4,5],
          "y":[4,3,4,3,3]}),
               pn.aes("x","y")
               )
     +pn.geom_point()
    )
ggp1.show(1)



ggp2=(
    pn.ggplot(pd.DataFrame({"x":[1,2,3,4,5],
                            "y":[4,3,4,3,3]}),
              pn.aes("x","y"))
    +pn.geom_point(color="red")
)
ggp2.show(2)


(ggp1+pn.geom_line(color='blue')).show(1) 
```


fkgruber:
I push one more change to clear the figure when you provide a figure number.

has2k1:
Thanks for this feature. I am not yet sure about the internal API changes. The external 
API `p.show(num=1)` is fine okay, but we should find a way not to add the parameter to 4 other internal functions i.e.

```python
ggplot.draw(bool=False, num=None)
facet.setup(num=None)
facet.make_figure(num=None)
facet._make_figure(num=None)
```

I will find a way to get this in before the next release.

has2k1:
Out of precaution because the API may change, I will look to add reusing figures as part of `v0.15.0` instead of `v0.14.0` which is the next one.  `v0.15.0` will do more things around figures.

Other options to consider for this functionality are:

```python
ggplot(figure_num=1)
```

```python
from plotnine.options import set_option
set_option(figure_num=1)
...
```

fkgruber:
Hi
Is there a branch for 0.15 that implements this feature? Would love to try

Issue #844, opened 2024-07-19
Wrong type hint for aes.__init__


nwweber:
Hey!

Issue: `aes.__init(...)` is type-hinted via its docstring. `**kwargs` is type-hinted incorrectly, viz.:

```
    **kwargs : dict
        Other aesthetic mappings
```

This type hint means 'the value of any named parameter should be of type `dict`'. Clearly, that is not the intention - `kwargs` _itself_ is a dict, the _values_ contained in it can be of arbitrary type.

This causes issues e.g. in PyCharm, which flags an incorrect type every single time I use `aes(...)` with keyword arguments.

Fix: change the docstring above to

```
    **kwargs : Any
        Other aesthetic mappings
```

PR #845, opened 2024-07-19
change type hint from dict to Any for **kwargs in aes.init

nwweber:
see https://github.com/has2k1/plotnine/issues/844


has2k1:
@nwweber, thanks.

PR #871, opened 2024-09-10
first attempt at strip_text_x ha adjustment for facets

jwhendy:
For review. Current status:

- hard codes `info.ha` as e.g. `'right'` for testing purposes
- accepts 'left', 'center', and 'right' text args as well as float (matched approach [here](https://github.com/has2k1/plotnine/blob/4159f9548fa5f95c02366efc324e88f9ba79af09/plotnine/_mpl/_plotnine_tight_layout.py#L193-L240))
- `anchor` (justification) variable added and set to match `info.ha` (if str passed), or is 'left' if `ha < 0.5` else 'right'

### DIFF
--- a/plotnine/_mpl/text.py
+++ b/plotnine/_mpl/text.py
@@ -63,14 +63,31 @@ class StripText(Text):
         self.patch.set_transform(info.ax.transAxes)
         self.patch.set_mutation_scale(0)
 
+        info.ha = "right"
+        margin = 0.05
+
         # Put text in center of patch
-        self._x = l + w / 2
+        if isinstance(info.ha, str):
+            lookup = {
+                "left": 0.0,
+                "center": 0.5,
+                "right": 1.0,
+            }
+            f = lookup[info.ha]
+            anchor = info.ha
+            if anchor == "right":
+                margin = -margin
+        else:
+            f = info.ha
+            anchor = "left" if info.ha < 0.5 else "right"
+
+        self._x = l + (w * f) + (w * margin)
         self._y = b + h / 2
 
         # "anchor" aligns before rotation so the right-strip get properly
         # centered text
         self.set_rotation_mode("anchor")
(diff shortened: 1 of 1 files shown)


has2k1:
`ha` and `va` get their values from

https://github.com/has2k1/plotnine/blob/4159f9548fa5f95c02366efc324e88f9ba79af09/plotnine/facets/strips.py#L70

You can read those from the theme to get rid of the hard-corded values.

```python
# This says, read the ha property of the strip_text_x themeable.
# If none has been set, use "center"
ha = theme.getp(("strip_text_x", "ha"), "center")
va = theme.getp(("strip_text_x", "va"), "bottom")
```

has2k1:
Ignore the margin for now.

Then to make the [0, 1] range seem natural when used for positioning we have to take more control of the alignment.

Note that, when:

1. `ha = "left"` or `ha = 0`, the left of the text should be at the left of the strip box. i.e. we set horizontal alignment to `left`.
2. `ha = "center" or `ha = 0.5`, the center of the text should be at the center of the strip box. i.e. we set horizontal alignment to `center`.
3. `ha = "right"` or `ha = 0`, the right of the text should be at the right of the strip box. i.e. we set horizontal alignment to `right`.

That means when;

- `ha = 0.2`, the `0.2 * width_of_text` point should be at `0.2 * width_of_strip_box`.
- `ha = 0.65`, the `0.65 * width_of_text` point should be at `0.65 * width_of_strip_box`.
- ...

The way to do this is set the horizontal alignment to the left and calculate the `x` position with that in mind. This is what happens here

https://github.com/has2k1/plotnine/blob/4159f9548fa5f95c02366efc324e88f9ba79af09/plotnine/_mpl/_plotnine_tight_layout.py#L211-L214

has2k1:
@jwhendy, Thanks for you effort on this issue, I have resolved it.

PR #922, opened 2025-03-11
Refactor reference: Add missing functions and remove non-exported ones

jeroenjanssens:


plotnine/__init__.py
has2k1, line 1: You should edit `plotnine._utils.dev.get_plotnine_all` as well, then use it. See the comment on lines 12-16.

PR #925, opened 2025-03-19
ux: check facet_*(scales) for valid values

TyberiusPrime:
I have repeatedly had to repeat plots because I can't keep it straight whether it's 'free_y' or 'freey' in a facet's scales, and plotnine doesn't complain about
invalid values in the scales argument.

Modern type checking does catch that, but 
not for example when you're writing a notebook.

This tiny fix just introduces a ValueError if scales argument isn't one of the supported values.


plotnine/facets/facet.py
```diff
@@ -121,6 +121,9 @@ def __init__(
         self.as_table = as_table
         self.drop = drop
         self.dir = dir
+        allowed_scales = ["fixed", "free", "free_x", "free_y"]
+        if scales not in allowed_scales:
+            raise ValueError(f"Scales must be one of {allowed_scales}")
```
has2k1, line 126: `scales` lowercase.

has2k1:
Thank you.

PR #969, opened 2025-08-12
Prefer `_repr_mimebundle_` for notebook display

manzt:
These changes replace `_ipython_display_` with `_repr_mimebundle_` protocol displaying plots within notebooks. Previously, plotnine used `_ipython_display_()` for display, which works via side effects and assumes an IPython/Jupyter environment.

Since IPython 6.1 (May 2017), the `_repr_mimebundle_()` protocol allows objects to return display data instead of triggering side effects. This decouples rendering from IPython, letting any compatible frontend (e.g., Jupyter, marimo) handle presentation without requiring the object to be aware of the runtime.

The repr continues to generate a single format (PNG, SVG, etc.) based on configuration to match prior behavior with the `_ipython_display_`.

References:
- https://ipython.readthedocs.io/en/stable/config/integrating.html
- https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.formatters.html


has2k1:
Thanks for this. I had not picked up on the precedence of `_repr_mimebundle_` and also to support both printing with side-effects (where necessary) and otherwise without them.

Issue #10, opened 2017-05-28
What about polar coords?


leiserfg:
I need a pie, but without polar coords I can't _cook it_.

has2k1:
Though I am not found of them, they will probably be in next minor version.

has2k1:
The plan was to do it together with [geom_carto](https://github.com/has2k1/plotnine/issues/14). Since both require working with projections of the coordinate system, I wanted to make sure that the code is refactored in a way that makes both possible.

When I last tried to contribute to [carto](https://github.com/SciTools/cartopy/pull/910), it moved quite slowly that the discussions/contributions would get reviews/responses after a  very long time, during which my context of the issue.

csoulette:
Any updates on this functionality? Or suggestions to achieve it? Ty~

CarlosGrohmann:
+1 for polar coords!

jgbradley1:
Did cartopy ever adopt the changes necessary for plotnine to implement polar coordinates?

larmstrong:
@jgbradley1 - Doesn't look like it so far. If it does, I'm not aware of it. (Neither is the code I'm currently writing. 🤓 )

hemantrattey:
Hi any updates on coord_polar being in plotnine or not?

nshahpazov:
Are there coord_polar?

bmwil3:
Ice at the poles might melt before we get coord_polar :(

has2k1:
> Ice at the poles might melt before we get coord_polar :(

Pretty optimistic! The ballpark is around the time the sun red-giants into earth's orbit.

MiguelSFeles:
hello any updates on coord_polar being in plotnine?

titi2338432:
coord_poloar  +1

dhimmel:
Would love `coord_polar`. Lot's of scientific applications besides pie charts. Am currently trying to plot a distribution of slope aspects (e.g. compass directions).

shaggycamel:
hello - May 2024. Still itching for coord_polar

jy-andy-hu:
May I ask what's the verdict on this topic? Is coord_polar coming ever or is there a different way to create pie charts with plotnine? 

I am trying to help people adopting plotnine, but not being able to create pie charts easily make it a hard sale...

iangow:
[...]

PR #1059, opened 2026-05-04
Added two new coordinate systems: `coord_polar` and `coord_radial`.

iangow:
This picks up on a thread started with #10. I implemented both the superseded (in `ggplot2`) `coord_polar()` and the newer  `coord_radial()`. Almost all arguments of the R equivalents have been implemented (I omitted `clip` because it is not supported in `coord_cartesian()` here either).

I created a small gallery of examples [here](https://iangow.github.io/coord-polar-test/), including examples from the `ggplot2` documentation and some interesting plots that seem to provide some rationale for using these coordinate systems.

I made some documentation mirroring the style of other plotnine documentation (and the original from `ggplot2`).

I got a lot of help from Claude Code (and Codex when I hit limits on Claude) on this, but I was careful to nudge it use Matplotlib's native `PolarAxes` as much as possible to keep the implementation lean. Looking at the code, it seems pretty concise.

Let me know if there's anything you like me to do to refine this or explain things better.


has2k1:
> I got a lot of help from Claude Code (and Codex when I hit limits on Claude) on this, but I was careful to nudge it use Matplotlib's native PolarAxes as much as possible to keep the implementation lean. Looking at the code, it seems pretty concise.

Using AI is fine as is any other tool, but for common development work I would rather leave out the attribution in the commit message. Mainly because we (the humans) are still responsible for reviewing, understanding and maintaining the code. The commit history should reflect intent, decisions and context.

We can include tools in the history when their output is deterministic, or when there is purpose in signalling some level of detachment from the result.

iangow:
> > I got a lot of help from Claude Code (and Codex when I hit limits on Claude) on this, but I was careful to nudge it use Matplotlib's native PolarAxes as much as possible to keep the implementation lean. Looking at the code, it seems pretty concise.
> 
> Using AI is fine as is any other tool, but for common development work I would rather leave out the attribution in the commit message. Mainly because we (the humans) are still responsible for reviewing, understanding and maintaining the code. The commit history should reflect intent, decisions and context.
> 
> We can include tools in the history when their output is deterministic, or when there is purpose in signalling some level of detachment from the result.

Sure. Would you like me to edit and resubmit?

iangow:
@has2k1 I added some tests and redid the commit messages to remove attribution to AI. Let me know if you want additional tests (e.g., output image comparisons).

has2k1:
> @has2k1 I added some tests and redid the commit messages to remove attribution to AI. Let me know if you want additional tests (e.g., output image comparisons).

I will have a better idea when I start reviewing it, hopefully next week.

iangow:
I realised that one other related thing from `ggplot2` that I did not implement is [`guide_axis_theta()`](https://ggplot2.tidyverse.org/reference/guide_axis_theta.html?q=coord_radial#null) (one thing this does is rotation of axis labels). It seems this wouldn't be too difficult, but I figure this can wait for a review of what is there now.

In the meantime, I may just work on cleaning up the [gallery of examples](https://iangow.github.io/coord-polar-test), as these may assist you [@has2k1] in your review.

has2k1:
@iangow, please rebase onto main if you can. Or I can take over.

iangow:
> @iangow, please rebase onto main if you can. Or I can take over.

@has2k1 Done.

plotnine/facets/facet.py
```diff
@@ -397,9 +398,16 @@ def _make_axes(self) -> tuple[p9GridSpec, list[Axes]]:
         gs = self._make_gridspec()
 
         # Create axes
+        from ..coords.coord_polar import coord_polar
+
+        subplot_kw = (
+            {"projection": "polar"}
+            if isinstance(self.plot.coordinates, coord_polar)
+            else {}
+        )
```
has2k1, line 354: For generality, the projection should be a property of `Coord`.


```python
class coord:
    _projection: str | None = None
    ...

class coord_polar:
    _projection = "polar"
    ...
````

File: plotnine/facets/facet.py [hunk 2]
```diff
@@ -355,6 +355,7 @@ def _inf_to_none(
 
         ax.tick_params(axis="x", which="major", pad=pad_x)
         ax.tick_params(axis="y", which="major", pad=pad_y)
+        self.coordinates.post_setup_ax(ax)
```
has2k1: We should move the functionality of `facet.set_limits_breaks_and_labels` to the `coord` class, maybe as `coord.setup_ax`. That is a better fit and the subclasses of `coord` would have the flexibility to completely override it or `super()` and add on.

plotnine/ggplot.py
```diff
@@ -520,6 +520,11 @@ def _draw_panel_borders(self):
         # grid lines below the borders. We leave ax.patch for the
         # background only.
         if self.theme.T.is_blank("panel_border"):
+            # For PolarAxes the default circular spine is separate from the
+            # panel border Rectangle; hide it explicitly when blank.
+            for ax in self.axs:
+                if "polar" in ax.spines:
+                    ax.spines["polar"].set_visible(False)
```
has2k1: Preferably, the blanking should be in `themes/themeable.py`. I think this would work.

```python
class panel_border:
    ...
    def blank_ax(self, ax):
        if "polar" in ax.spines:
            ax.spines["polar"].set_visible(False)
```

has2k1: Custom `coord` classes that do not ship with plotnine and cannot modify the `themeables` could do this kind of manipulation in `coord.setup_ax`.

Later, we could consider making the `themeables` extendable!

plotnine/facets/facet.py
```diff
@@ -397,9 +398,16 @@ def _make_axes(self) -> tuple[p9GridSpec, list[Axes]]:
...
+
+        subplot_kw = (
+            {"projection": "polar"}
+            if isinstance(self.plot.coordinates, coord_polar)
+            else {}
+        )
         it = itertools.product(range(self.nrow), range(self.ncol))
         for i, (row, col) in enumerate(it):
-            axsarr[row, col] = self.figure.add_subplot(gs[i])
+            axsarr[row, col] = self.figure.add_subplot(gs[i], **subplot_kw)
```
has2k1: With the above change this would be to

```python
axsarr[row, col] = self.figure.add_subplot(
    gs[i],
    projection=self.plot.coordinates._projection
)
```

plotnine/coords/coord.py
```diff
@@ -104,6 +104,22 @@ def aspect(self, panel_params: panel_view) -> float | None:
...
+        """
+
+    def post_setup_ax(self, ax: Any) -> None:
+        """
+        Hook called for each axes after set_limits_breaks_and_labels.
+
+        Override in subclasses to apply per-axes settings that must
+        run after the facet has set tick positions and label padding.
+        """
+
```
has2k1, line 348: With `coord.setup_ax` (see below), this method would be unnecessary.

has2k1 (changes requested):
For the first pass, this is mainly a refactoring. The goal is to limit the external footprint of custom `coord` classes outside their modules.

plotnine/ggplot.py
iangow: @has2k1 Let me know if you want help with any of the above.

has2k1: You can fix the issues.

iangow:
@has2k1 I took a stab at the requested changes (I made an issue for each of the four changes you requested and you can see the associated commits above). I also implemented the `guide_axis_theta()` that was missing. I updated the page [here](https://iangow.github.io/coord-polar-test/) to check the changes and to demo the new `guide_axis_theta()`.

has2k1:
On a follow up review, I discovered that while the geoms are drawn properly, fixing the issues with gridlines, ticks and labels some bigger changes. I'm working on that and will finish up this PR.

has2k1:
One of the sticky requirements for radial coordinates is the ability to have theta axis ticks on the outside (along the circumference). Translated to cartesian coordinates, this is the opposite side (e.g. x axis at top of the panel) and PR #1085 makes it possible.

Also, #1085 starts of with cherry-picking a [refactor](https://github.com/has2k1/plotnine/pull/1059/commits/7a65e2803a51161c3f4b78a95793b2b04ca67abd) request in the earlier review.

has2k1:
@iangow thanks for kicking this off. I am pushing it out in the next pre-release v0.16.0a12.

iangow:
> @iangow thanks for kicking this off. I am pushing it out in the next pre-release v0.16.0a12.

I updated the "demo" page [here](https://iangow.github.io/coord-polar-test/). Though it seems that `guide_axis_theta()` dropped out somewhere along the way, so the page for that doesn't do much.

has2k1:
> Though it seems that `guide_axis_theta()` dropped out somewhere along the way, so the page for that doesn't do much.

`guide_axis_theta()` came before the more basic `guide_axis()`. As the polar implementation got more involved it got in the way before I could be sure of how to integrate it. But it will common either before the release of soon after.

PR #1083, opened 2026-06-18
Removed copy call from mpl.rcParams for params update when using defaults.

### LINKED ISSUE #1071: Plotting with Matplotlib theme gives error
I created a plot and tried to change the theme to a Matplotlib style. When I tried to draw the figure, the following error occurred.

AttributeError: 'object' object has no attribute 'startswith'

I also found that the code ran successfully when I ran it with a debugger.

### Example
```
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "matplotlib",
#     "plotnine",
#     "polars",
#     "pyarrow",
# ]
# ///
from plotnine import *
import polars as pl


def plot_temp_data(df):
    """ """
    return (
        ggplot(data=df, mapping=aes(x="x", y="y"))
        + geom_line()
        + labs(x="x", y="y")
        + theme_matplotlib()
    )

def main():
    length = 10
    df = pl.DataFrame({'x': [idx for idx in range(length)], 'y': [2 for _ in range(length)]})
    temp_fig = plot_temp_data(df)
    temp_fig.draw(show=True)

if __name__ == '__main__':
    main()
```

zhymirt:
The copy method for Matplotlib's rcParams creates a raw copy without resolving the backend. Calling update directly on the default rcParams causes the backend to resolve before copying the value over.

Closes #1071

### DIFF
--- a/plotnine/themes/theme_matplotlib.py
+++ b/plotnine/themes/theme_matplotlib.py
@@ -136,7 +136,7 @@ class theme_matplotlib(theme):
         )
 
         if use_defaults:
-            self._rcParams.update(mpl.rcParams.copy())
+            self._rcParams.update(mpl.rcParams)
 
         if fname:
             self._rcParams.update(mpl.rc_params_from_file(fname))


has2k1:
Thank you @zhymirt.

PR #1085, opened 2026-06-23
Support moving axes to the opposite side of the panel

has2k1:
Adds the ability to place a position scale's axis on the opposite side of the panel — the x axis on top, the y axis on the right — together with per-side theming so each axis side can be styled independently.

```python
  from plotnine import *

  (
      ggplot(mtcars, aes("wt", "mpg"))
      + geom_point()
      + scale_x_continuous(position="top")
      + scale_y_continuous(position="right")
  )
```


TyberiusPrime:
Will this eventually lead to dual axis plots?

has2k1:
> Will this eventually lead to dual axis plots?

A big yes. And not eventually, rather immediately!

PR #1087, opened 2026-06-25
Fix: missing f-marker on facet-scales-error-message

TyberiusPrime:
Noticed this error message literally printing "Argument `scales` must be one of {allowed_scales}."',
instead of stamping in the allowed values.

Pretty sure I messed that up last year, my apologies.


has2k1:
Thank you. Though, I kind of recall the linter being able to flag that issue.

PR #1096, opened 2026-07-04
fix(layer): cast polars decimal columns to float

arpitjain099:
A polars Decimal column comes through to_pandas() as a pandas object column of decimal.Decimal values, so plotnine reads it as discrete and a continuous scale rejects it with "Discrete value supplied to continuous scale" (#1033). This casts those Decimal object columns to float right after the conversion, following the cheap-hack approach you mentioned in the issue, so such a column now behaves like any other numeric one instead of needing a manual cast to Float. I added a small regression test that builds a plot from a Decimal column on a continuous scale. Thanks!

### DIFF
--- a/plotnine/layer.py
+++ b/plotnine/layer.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import decimal
 import typing
 from copy import copy, deepcopy
 from typing import Iterable, List, cast, overload
@@ -29,6 +30,23 @@ if typing.TYPE_CHECKING:
         DataLike,
         LayerDataLike,
     )
+
+
+def _convert_decimal_columns(data: pd.DataFrame) -> pd.DataFrame:
+    """
+    Cast columns of decimal.Decimal values to float
+
+    polars' `to_pandas()` maps a Decimal column onto a pandas object column
+    holding `decimal.Decimal` values. plotnine treats object columns as
+    discrete, so such a column cannot be used with a continuous scale. Casting
+    it to float lets it behave like any other numeric column.
+    """
+    for col in data.columns:
+        if data[col].dtype == object:
+            non_null = data[col].dropna()
+            if len(non_null) and isinstance(non_null.iloc[0], decimal.Decimal):
+                data[col] = data[col].astype("float64")
+    return data
 
 
 class layer:
@@ -222,7 +240,9 @@ class layer:
         if plot_data is None:
             data = pd.DataFrame()
         elif hasattr(plot_data, "to_pandas"):
-            data = cast("DataFrameConvertible", plot_data).to_pandas()
+            data = _convert_decimal_columns(
+                cast("DataFrameConvertible", plot_data).to_pandas()
+            )
         else:
             data = cast("pd.DataFrame", plot_data)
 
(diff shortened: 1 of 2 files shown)


plotnine/layer.py
```diff
@@ -31,6 +32,23 @@
     )
 
 
+def _convert_decimal_columns(data: pd.DataFrame) -> pd.DataFrame:
```
has2k1: Rename to `_decimal_columns_to_float` and put function at the bottom of the module.

plotnine/layer.py
```diff
@@ -31,6 +32,23 @@
...
+    holding `decimal.Decimal` values. plotnine treats object columns as
+    discrete, so such a column cannot be used with a continuous scale. Casting
+    it to float lets it behave like any other numeric column.
+    """
+    for col in data.columns:
+        if data[col].dtype == object:
+            non_null = data[col].dropna()
+            if len(non_null) and isinstance(non_null.iloc[0], decimal.Decimal):
+                data[col] = data[col].astype("float64")
+    return data
```
has2k1: As this function should run for all dataframes, we are have to mind expensive the `.dropna()` which builds a filtered copy of the column. This holds for pandas 2 and pandas 3 (without pyarrow) installed and the fact that most object columns would be string columns! We should use a mask, `.notna()`.

```python
def _decimal_columns_to_float(data):
    for col in data.columns:
        series = data[col]
        if series.dtype == object:
            mask = series.notna()
            if mask.any() and isinstance(series[mask.idxmax()], decimal.Decimal):
                data[col] = series.astype("float64")
    return data
```

plotnine/layer.py
```diff
@@ -249,9 +269,9 @@ def _make_layer_data(self, plot_data: DataLike | None):
...
             if hasattr(self._data, "to_pandas"):
-                self.data = cast(
-                    "DataFrameConvertible", self._data
-                ).to_pandas()
+                self.data = _convert_decimal_columns(
+                    cast("DataFrameConvertible", self._data).to_pandas()
+                )
             elif isinstance(self._data, pd.DataFrame):
                 self.data = self._data.copy()
             else:
```
has2k1, line 258: Finally the solution is applied to every dataframe.

```python
self.data = _decimal_columns_to_float(self.data)
```

tests/test_ggplot_internals.py
```diff
@@ -365,6 +367,29 @@ def to_pandas(self):
...
+    from decimal import Decimal
+
+    class DecimalData:
+        def to_pandas(self):
+            return pd.DataFrame(
+                {
+                    "x": ["a", "b", "c"],
+                    "y": [Decimal(1), Decimal(2), Decimal(3)],
+                }
+            )
```
has2k1: With the revised solution you can use a straight pandas frame.

has2k1 (changes requested):
The root cause is pandas frames with Decimal columns, so we should fix that.

arpitjain099:
Thanks, that makes sense. I moved the fix to the root: `self.data = _decimal_columns_to_float(self.data)` now runs once at the end of `_make_layer_data`, so every dataframe is covered and the per-call-site casts on the polars paths are gone. I renamed the helper, put it at the bottom of the module, and switched to the `notna()` mask so we don't pay for `dropna()` on every object column. The test is now a plain pandas frame with a Decimal column, and the full test file passes.

has2k1:
@arpitjain099, thank you.

PR #1103, opened 2026-08-08
Fix typos in docstrings

latent-9:
A few small typos in docstrings:

- `plotnine/stats/stat_bin.py`, `stat_bin_2d.py`, `stat_summary_bin.py`: "to use bins bins that" -> "to use bins that".
- `plotnine/stats/stat_ellipse.py`: "A dictionary with with the following" -> "with the following"; and "the fitted covarince matrix" -> "covariance".
- `plotnine/themes/theme_light.py`: "light grey lines lines and axes" -> "lines and axes".
- `plotnine/themes/theme.py`: "override this method method as long as" -> "this method as long as".

Documentation only.

### DIFF
--- a/plotnine/stats/stat_bin.py
+++ b/plotnine/stats/stat_bin.py
@@ -25,7 +25,7 @@ class stat_bin(stat):
     ----------
     {common_parameters}
     binwidth : float, default=None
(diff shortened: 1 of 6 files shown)


has2k1:
Thank you.

PR #1119, opened 2026-09-03
fix(stat_summary): compute mean_se from the sample variance

dylanpulver:
`mean_se` builds the standard error from `np.var(series)`, whose default `ddof=0` is the **population** variance. The standard error of the mean uses the **sample** variance, so the interval is too narrow by `sqrt((n-1)/n)` — 29% at n=2, 5% at n=10.

ggplot2's reference implementation (`R/stat-summary.R`):

```r
se <- mult * sqrt(stats::var(x) / length(x))
```

R's `stats::var` is the sample variance.

The two sibling summary functions in this module already use sample estimators — `mean_sdl` via `Series.std()` and `mean_cl_normal` via `scipy.stats.sem`, both `ddof=1`. `mean_se` was the only one on `ddof=0`.

Measured half-width for `[2,4,4,4,5,5,7,9]` (n=8):

| | half-width |
|---|---|
| before | 0.7071067812 |
| `s/sqrt(n)` = `scipy.stats.sem` = ggplot2 | 0.7559289460 |

ratio 0.9354143467 = `sqrt(7/8)` exactly.

The fix keeps ggplot2's `sqrt(var/n)` shape. `series.std()/sqrt(n)` also passes the suite, but is worse: `ndarray.std()` defaults to `ddof=0`, so it would keep the bug for array input and raise on a list.

`mean_se` is the default `fun_data` for `stat_summary_bin`, so `mean_se.png` is regenerated (via `tools/update_baseline_images.py`, then `oxipng -o 4 --strip safe`).

Tests: 850 passed, 5 skipped, 1 xpassed, 0 failed. An earlier version of this description reported 6 `test_geom_smooth` failures; those were missing dependencies in my checkout. The added test fails on the unpatched function.

Found and drafted with Claude Opus 5 (`claude-opus-5`); I reviewed and verified the change.


tests/test_stat_summary.py
```diff
@@ -50,6 +51,23 @@ def test_mean_se():
...
+    assert res["y"].iloc[0] == pytest.approx(s.mean())
+    assert res["ymax"].iloc[0] - res["y"].iloc[0] == pytest.approx(expected)
+    assert res["y"].iloc[0] - res["ymin"].iloc[0] == pytest.approx(expected)
+
+    # mult scales the interval
+    res2 = mean_se(s, mult=2)
+    half_width = res2["ymax"].iloc[0] - res2["y"].iloc[0]
+    assert half_width == pytest.approx(2 * expected)
+
+
```
has2k1: We can remove this test. The comment in the function and the visual test are sufficient.

has2k1 (changes requested):
Thanks for catching that. For now I [prefer](https://github.com/has2k1/plotnine/pull/1059#issuecomment-4378001016) no AI attribution in the commit message.

has2k1:
@dylanpulver, thank you.

PR #1125, opened 2026-09-17
Fix docstrings that name parameters the operator methods do not have

VenishPaneliya:
Four of the `+` operator methods document a parameter name their signature doesn't use, so following the docs and passing that keyword raises `TypeError`.

| file | method | signature | documented |
|---|---|---|---|
| `ggplot.py` | `ggplot.__add__` | `rhs` | `other` |
| `geoms/geom.py` | `geom.__radd__` | `other` | `plot` |
| `guides/guides.py` | `guides.__radd__` | `other` | `plot` |
| `stats/stat.py` | `stat.__radd__` | `other` | `gg` |

`other` is the established name here. Of the 24 `__radd__` methods in the package, the two that document their parameter at all — `ggplot.__radd__` and `composition/_types.py` — both use `other`; the remaining 19 have no Parameters section.

`ggplot.py` makes the case on its own: its `__radd__` documents `other` correctly, while a few lines away `__add__` also says `other` for a parameter that is actually called `rhs`.

I left `stat.__radd__`'s `out :` entry alone — that's the name under `Returns`, which is valid numpydoc, not a mistake.

Docs only, no behaviour change. `ruff check` and `ruff format --check` are clean on all four files.


VenishPaneliya:
Note on the red `typecheck (3.14)`: both errors are in `plotnine/layer.py:879` —

```
plotnine/layer.py:879:17 - error: No overloads for "__getitem__" match the provided arguments
plotnine/layer.py:879:17 - error: Argument of type "Hashable" cannot be assigned to parameter "idx" ...
```

`layer.py` isn't in this PR — it only touches `ggplot.py`, `geoms/geom.py`, `guides/guides.py` and `stats/stat.py`, and only docstring text in each. This looks like pandas-stubs drift on 3.14 rather than anything from here. Happy to rebase if you'd like a clean run once that's sorted.

has2k1:
@VenishPaneliya, thank you. I fixed the typechecking.

